PR #14, opened 2022-10-08
Deprovisioning

ukff:
**Description**
This PR introduces a new feature - deprovisioning.


docs/deprovisioning.md
```diff
@@ -0,0 +1,18 @@
...
+
+```
+kubectl delete btpoperator yourbtpoperator
+```
+
+This command will trigger deletion of all service bindings, service instances and btp operator resources on your cluster.
+
+At first, the deletion of deployment and webhooks which yourbtpoperator manages will happen, then deprovisioning flow will try to perform deletion in hard delete mode, which mean it will find all service bindings(and their secrets) and service instances across all namespaces and will try to make kubernetes delete on it.
+The time limit for this operation is 20 minutes.
+After this time, or in case of error in hard deletion, the system will go into soft delete mode which mean deletion of finalizers from service instances and bindings.
```
IwonaLanger, line 13:
```suggestion
After this time, or in case of an error in hard deletion, the system goes into soft delete mode, which runs deletion of finalizers from service instances and bindings.
```

docs/deprovisioning.md
```diff
@@ -0,0 +1,18 @@
...
+Regardless of mode, at later step, the all btp operator resources marked with label "managed-by:btp-operator" are deleted.
```
IwonaLanger, line 14:
```suggestion
Regardless of mode, in the next step, the all BtpOperator resources marked with the "managed-by:btp-operator" label are deleted.
```

docs/deprovisioning.md
```diff
@@ -0,0 +1,18 @@
...
+If proccess pass with success then finally finalizer on yourbtpoperator itself is removed and the resource is deleted.
```
IwonaLanger, line 15:
```suggestion
If the process runs successfully, the finalizer on `yourbtpoperator` itself is removed and the resource is deleted.
```

docs/deprovisioning.md
```diff
@@ -0,0 +1,18 @@
...
+If any error will happen during deprovisioning, then yourbtpoperator will be set to Error state.
```
IwonaLanger, line 16:
```suggestion
If an error occurs during deprovisioning, `yourbtpoperator` is set to `Error`.
```

operator/module-chart/README.md
```diff
@@ -108,7 +108,7 @@ caBundle: {{ b64enc $ca.Cert }}
 
 ### Add sap-btp-operator labels
 
-The deployment and service must contain btp operator specific labels (deployment spec, deployment template and the service label selector):
+The deployment and service must contain btp operator specific labels (deployment spec, deployment template and the service labels selector):
```
IwonaLanger, line 111:
```suggestion
The deployment and service must contain btp-operator specific labels (deployment spec, deployment template and the service labels selector):
```

docs/deprovisioning.md
```diff
@@ -0,0 +1,18 @@
...
+
+![KEB diagram](./assets/keb-architecture.svg)
```
IwonaLanger, line 18:
```suggestion
![Deprovisioning diagram](./assets/deprovisioning.svg)
```

IwonaLanger, line 18: In the diagram, is it `Reconcile` or `Reconciler`?

docs/deprovisioning.md
```diff
@@ -0,0 +1,18 @@
+## Deprovisioning
+
+To start the deprovisioning process, use the following command:
+
+```
+kubectl delete btpoperator yourbtpoperator
```
IwonaLanger, line 6:
```suggestion
kubectl delete btpoperator your-btpoperator
```

docs/deprovisioning.md
```diff
@@ -0,0 +1,18 @@
...
+```
+
+The command triggers deletion of all service bindings, service instances and BtpOperator resources on your cluster.
+
+After the deletion of deployment and webhooks which `yourbtpoperator` manages, the deprovisioning flow tries to perform deletion in hard delete mode. When it finds all service bindings and their secrets, and service instances across all namespaces, it tries to delete Kubernetes.
```
IwonaLanger, line 11:
```suggestion
After the deletion of deployment and webhooks which `your-btpoperator` manages, the deprovisioning flow tries to perform deletion in hard delete mode. When it finds all service bindings and their secrets, and service instances across all namespaces, it tries to delete Kubernetes.
```

docs/deprovisioning.md
```diff
@@ -0,0 +1,18 @@
...
+The time limit for this operation is 20 minutes.
+After this time, or in case of an error in hard deletion, the system goes into soft delete mode, which runs deletion of finalizers from service instances and bindings.
+Regardless of mode, in the next step, the all BtpOperator resources marked with the "managed-by:btp-operator" label are deleted.
+If the process runs successfully, the finalizer on `yourbtpoperator` itself is removed and the resource is deleted.
+If an error occurs during deprovisioning, `yourbtpoperator` is set to `Error`.
```
IwonaLanger, line 16:
```suggestion
If the process runs successfully, the finalizer on `your-btpoperator` itself is removed and the resource is deleted.
If an error occurs during deprovisioning, `your-btpoperator` is set to `Error`.
```

docs/deprovisioning.md
```diff
@@ -0,0 +1,18 @@
...
+```
+kubectl delete btpoperator your-btpoperator
+```
+
+The command triggers deletion of all service bindings, service instances and BtpOperator resources on your cluster.
+
+After the deletion of deployment and webhooks which `your-btpoperator` manages, the deprovisioning flow tries to perform deletion in hard delete mode. When it finds all service bindings and their secrets, and service instances across all namespaces, it tries to delete them.
+The time limit for this operation is 20 minutes.
+After this time, or in case of an error in hard deletion, the system goes into soft delete mode, which runs deletion of finalizers from service instances and bindings.
+Regardless of mode, in the next step, the all BtpOperator resources marked with the "managed-by:btp-operator" label are deleted.
```
IwonaLanger, line 14:
```suggestion
Regardless of mode, in the next step, all the BtpOperator resources marked with the "managed-by:btp-operator" label are deleted.
```

IwonaLanger (approved):

operator/controllers/btpoperator_controller.go
```diff
@@ -19,46 +16,111 @@ package controllers
...
-	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 const (
-	chartPath         = "./module-chart"
+	chartPath         = "/Users/lj/Go/src/github.com/kyma-project/modularization/btp-manager/operator/module-chart"
```
szwedm, line 50: avoid using absolute hardcoded paths

operator/controllers/btpoperator_controller.go
```diff
@@ -19,46 +16,111 @@ package controllers
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/rest"
 	"reflect"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 	"time"
```
szwedm, line 33: it's good practice to sort imports

operator/controllers/btpoperator_controller.go
```diff
@@ -89,27 +154,51 @@ func (r *BtpOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	}
+	requests := make([]reconcile.Request, len(btpOperators.Items))
+	for i, item := range btpOperators.Items {
+		requests[i] = reconcile.Request{NamespacedName: k8sgenerictypes.NamespacedName{Name: item.GetName(), Namespace: item.GetNamespace()}}
+	}
+
+	return requests
+}
 
+func (r *BtpOperatorReconciler) SetStatus(new types.State, ctx context.Context, cr *v1alpha1.BtpOperator) error {
```
szwedm, line 189: We can shorten this method to two lines of code:
cr.SetStatus(status.WithState(newState))
return r.Status().Update(ctx, cr)
and call it UpdateBtpOperatorStatus()

operator/controllers/btpoperator_controller.go
```diff
@@ -143,17 +232,37 @@ func (r *BtpOperatorReconciler) HandleProcessingState(ctx context.Context, cr *v
...
+
+	return r.SetStatus(types.StateProcessing, ctx, cr)
+}
+
+func (r *BtpOperatorReconciler) HandleDeletingState(ctx context.Context, cr *v1alpha1.BtpOperator) error {
+	logger := log.FromContext(ctx)
+	if err := r.handleDeprovisioning(ctx); err != nil {
+		logger.Error(err, "deprovisioning failed")
+		return r.SetStatus(types.StateError, ctx, cr)
+	} else {
```
szwedm, line 256: we can skip `else` and treat actions in its code block as default workflow when there is no error from handleDeprovisioning()

operator/controllers/btpoperator_controller.go
```diff
@@ -143,17 +232,37 @@ func (r *BtpOperatorReconciler) HandleProcessingState(ctx context.Context, cr *v
...
+		logger.Info("deprovisioning success. clearing finalizers for btp manager")
+		cr.SetFinalizers([]string{})
+		if err := r.Update(ctx, cr); err != nil {
+			return err
```
szwedm, line 260: maybe it's worth to provide more context to the error thrown from Update() using `fmt.Errorf()`

operator/controllers/btpoperator_controller.go
```diff
@@ -347,3 +417,283 @@ func (r *BtpOperatorReconciler) watchBtpOperatorUpdatePredicate() predicate.Func
 		},
 	}
 }
+
+func (r *BtpOperatorReconciler) handleDeprovisioning(ctx context.Context) error {
+	logger := log.FromContext(ctx)
+	logger.Info("btp operator is under deletion")
```
szwedm, line 423: how about `Deprovisioning BTP Operator`?

operator/controllers/btpoperator_controller.go
```diff
@@ -143,17 +232,37 @@ func (r *BtpOperatorReconciler) HandleProcessingState(ctx context.Context, cr *v
...
+	logger.Info("Handling Error state")
+
+	return r.SetStatus(types.StateProcessing, ctx, cr)
+}
+
+func (r *BtpOperatorReconciler) HandleDeletingState(ctx context.Context, cr *v1alpha1.BtpOperator) error {
+	logger := log.FromContext(ctx)
+	if err := r.handleDeprovisioning(ctx); err != nil {
+		logger.Error(err, "deprovisioning failed")
+		return r.SetStatus(types.StateError, ctx, cr)
```
szwedm, line 255: the error thrown from handleDeprovisioning() is ignored when HandleDeletingState() returns, it's overshadowed by error from SetStatus() (r.Status().Update(ctx, cr))

operator/controllers/btpoperator_controller.go
```diff
@@ -347,3 +417,283 @@ func (r *BtpOperatorReconciler) watchBtpOperatorUpdatePredicate() predicate.Func
...
+}
+
+func (r *BtpOperatorReconciler) hardDelete(ctx context.Context, gvk schema.GroupVersionKind, namespaces *corev1.NamespaceList) error {
+	logger := log.FromContext(ctx)
+	object := &unstructured.Unstructured{}
+	object.SetGroupVersionKind(gvk)
+
+	for _, namespace := range namespaces.Items {
+		if err := r.DeleteAllOf(ctx, object, client.InNamespace(namespace.Name)); err != nil {
+			logger.Error(err, "Err while doing delete all of")
```
szwedm, line 512: how about changing the error to logger.Error(err, "while deleting all resources", "kind", object.GetKind())

operator/controllers/btpoperator_controller.go
```diff
@@ -347,3 +417,283 @@ func (r *BtpOperatorReconciler) watchBtpOperatorUpdatePredicate() predicate.Func
...
+	if err := r.hardDelete(ctx, bindingGvk, namespaces); err != nil {
+		logger.Error(err, "hard delete of binding failed")
+		anyErr = true
+	}
+	if err := r.hardDelete(ctx, instanceGvk, namespaces); err != nil {
+		logger.Error(err, "hard delete of instance failed")
+		anyErr = true
+	}
+
+	if r.reconcilerConfig.testFlag != "" {
```
szwedm, line 481: we can use fake k8s client and overshadow the methods to change client's behavior

operator/controllers/btpoperator_controller.go
```diff
@@ -347,3 +417,283 @@ func (r *BtpOperatorReconciler) watchBtpOperatorUpdatePredicate() predicate.Func
...
+		if err != nil {
+			logger.Error(err, "Checking failed")
+			success <- false
+			return
+		}
+		if !resourcesLeft {
+			success <- true
+			return
+		}
+		time.Sleep(retryInterval)
```
szwedm, line 501: this can potentially lead to endless loop

operator/controllers/btpoperator_controller.go
```diff
@@ -347,3 +417,283 @@ func (r *BtpOperatorReconciler) watchBtpOperatorUpdatePredicate() predicate.Func
...
+		err, instancesLeft := anyLeft(namespace.Name, instanceGvk)
+		if err != nil {
+			return err, true
+		}
+		err, bindingsLeft := anyLeft(namespace.Name, bindingGvk)
+		if err != nil {
+			return err, true
+		}
+		if instancesLeft == false && bindingsLeft == false {
+			return nil, false
```
szwedm, line 543: method can return after checking only one namespace

operator/controllers/btpoperator_controller.go
```diff
@@ -347,3 +417,283 @@ func (r *BtpOperatorReconciler) watchBtpOperatorUpdatePredicate() predicate.Func
...
+	if err := r.DeleteAllOf(ctx, validatingWebhook, labelFilter); err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (r *BtpOperatorReconciler) cleanUpAllBtpOperatorResources(ctx context.Context, namespaces *corev1.NamespaceList) error {
```
szwedm, line 640: collecting all GVKs from the cluster just to delete a few connected to the BTP Service Operator chart seems like overdo. Could GVKs be limited only to these present in the chart?

operator/controllers/btpoperator_controller.go
```diff
@@ -347,3 +417,283 @@ func (r *BtpOperatorReconciler) watchBtpOperatorUpdatePredicate() predicate.Func
...
+
+func (r *BtpOperatorReconciler) softDelete(ctx context.Context, gvk *schema.GroupVersionKind) error {
+	listGvk := *gvk
+	listGvk.Kind = gvk.Kind + "List"
+	list := &unstructured.UnstructuredList{}
+	list.SetGroupVersionKind(listGvk)
+
+	if err := r.List(ctx, list); err != nil {
+		err = fmt.Errorf("%w; could not list in soft delete", err)
+		return err
```
szwedm, line 580: can be replaced with one liner:
```
return fmt.Errorf("while listing resources: %w", err)
```

operator/controllers/btpoperator_controller.go
```diff
@@ -347,3 +417,283 @@ func (r *BtpOperatorReconciler) watchBtpOperatorUpdatePredicate() predicate.Func
...
+	}
+
+	return nil, true
+}
+
+func (r *BtpOperatorReconciler) handleSoftDelete(ctx context.Context, namespaces *corev1.NamespaceList) error {
+	logger := log.FromContext(ctx)
+	logger.Info("hard delete failed. trying to perform soft delete")
+
+	if err := r.softDelete(ctx, &instanceGvk); err != nil {
```
szwedm, line 554: first try to process service bindings, then service instances

operator/controllers/btpoperator_controller.go
```diff
@@ -347,3 +417,283 @@ func (r *BtpOperatorReconciler) watchBtpOperatorUpdatePredicate() predicate.Func
...
+			}
+		}
+	}
+
+	if err := r.List(ctx, list); err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+	} else {
+		if len(list.Items) > 0 {
```
szwedm, line 605: won't this condition be always true where there are service instances/service bindings in the cluster? Before this condition, only Secrets are deleted and Finalizers are removed from resources.

operator/controllers/btpoperator_controller.go
szwedm: main thread will continue after the timeout, but handleHardDelete() runs in a separate goroutine which can be in the loop, example:
https://go.dev/play/p/Qfm8mmX9j3H

operator/controllers/btpoperator_controller.go
szwedm: BTP Manager holds a local copy of a chart and we could fetch all the required GVKs from there. Let's discuss it f2f.

operator/controllers/btpoperator_controller.go
szwedm: what about the case when DeleteAllOf() fails and `anyErr` is set to true?

szwedm: I think we may run into problems when deletion timestamp is not set and we reach softDelete().

PR #42, opened 2022-11-30
Fix for passing wrong value into cluster.id + sap-btp-manager change …

PK85:
…was not triggering reconciliation

**Description**

- Fix for passing wrong value into cluster.id
- Fix for sap-btp-manager change is not triggering reconciliation
- ...

**Related issue(s)**


szwedm (approved):
/lgtm

PR #41, opened 2022-11-30
Introduce configurability via CLI parameters and ConfigMap

wozniakjan:
This PR introduces configurability for btp-manager via CLI parameters and `ConfigMap`

Example:
```yaml
apiVersion: v1
data:
  ProcessingStateRequeueInterval: 10s
kind: ConfigMap
metadata:
  creationTimestamp: "2022-12-01T14:02:14Z"
  name: sap-btp-manager
  namespace: kyma-system
  resourceVersion: "6476"
  uid: f6f522db-9b4c-4254-a7ae-7e306305544d
```

closes: https://github.com/kyma-project/btp-manager/issues/31

PR #61, opened 2022-12-12
Unify configuration of hard delete timeout

wozniakjan:
**Description**

PR https://github.com/kyma-project/btp-manager/pull/41 has introduced dynamic configurability of btp-manager timeouts, sync intervals and other basic settings. The hard delete timeout was implemented differently. 

This PR unifies the code to make the configuration of `HardDeleteTimeout` match the conventions with all other timeouts and intervals.

**Related issue(s)**

### DIFF
--- a/operator/controllers/btpoperator_controller.go
+++ b/operator/controllers/btpoperator_controller.go
@@ -62,6 +62,7 @@ var (
 	ReadyStateRequeueInterval      = time.Hour * 1
 	ReadyTimeout                   = time.Minute * 1
 	RetryInterval                  = time.Second * 10
+	HardDeleteTimeout              = time.Minute * 20
 )
 
 const (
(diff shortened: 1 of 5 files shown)


operator/controllers/btpoperator_controller.go
```diff
@@ -62,6 +62,7 @@ var (
 	ReadyStateRequeueInterval      = time.Hour * 1
 	ReadyTimeout                   = time.Minute * 1
 	RetryInterval                  = time.Second * 10
```
szwedm, line 64: How about changing `RetryInterval` var name to something like `HardDeleteCheckInterval` and passing it to `time.Sleep()` func in L661?

operator/controllers/btpoperator_controller.go
```diff
@@ -660,14 +658,14 @@ func (r *BtpOperatorReconciler) handleHardDelete(ctx context.Context, namespaces
 			return
 		}
 
-		time.Sleep(r.hardDeleteTimeout / 20)
+		time.Sleep(HardDeleteTimeout / 20)
```
szwedm, line 661: We can use var from L64 here to have configurable interval for hard delete check

operator/controllers/suite_test.go
```diff
@@ -96,7 +96,7 @@ var _ = BeforeSuite(func() {
 		WaitForChartReadiness: false,
 	}
 	k8sClientFromManager = k8sManager.GetClient()
-	reconciler.SetHardDeleteTimeout(hardDeleteTimeout)
+	HardDeleteTimeout = hardDeleteTimeout
```
szwedm, line 99: We also need to adjust hard delete check interval for tests. Default value is set too high for tests `hardDeleteTimeout`, so I suggest to set `HardDeleteCheckInterval` with `hardDeleteTimeout/20` value

szwedm (approved):
/lgtm

IwonaLanger (approved):

PR #62, opened 2022-12-13
WIP: End2end test p4

tillknuesting:
**Description**

- ...
- ...
- ...

**Related issue(s)**


testing/go/e2e_test.go
```diff
@@ -10,16 +10,54 @@ import (
...
 
 	c.Stdout = b
 	c.Run()
 	out := b.String()
 
 	log.Println(out)
 
-	if len(out) == 0 {
-		t.Error("No CRD found")
+	if !strings.Contains(out, "Running") {
```
PK85, line 23: Where do you expect to get "Running"? we do not have such string, see: https://github.com/kyma-project/btp-manager/blob/main/docs/operations.md#conditions

testing/go/e2e_test.go
```diff
@@ -10,16 +10,54 @@ import (
 func TestEndToEnd(t *testing.T) {
 	log.Println("Starting E2E test...")
 
-	log.Println("kubectl get crd")
-	c, b := exec.Command("kubectl", "get", "crd"), new(strings.Builder)
+	// Check that the BTPOperator resource is created and has the expected status.
+	log.Println("kubectl get BTPOperator")
```
PK85, line 14: Fisrt you need to provide dummy secret and prereq then check BTPOperator CR status

testing/go/e2e_test.go
```diff
@@ -10,16 +10,54 @@ import (
...
+	// Create a dummy serviceInstance using the dummy secret and check that it is in error state.
+	log.Println("kubectl create -f deployments/prerequisites.yaml")
+	c, b = exec.Command("kubectl", "create", "-f", "deployments/prerequisites.yaml"), new(strings.Builder)
+
+	c.Stdout = b
+	c.Run()
+	out = b.String()
+
+	log.Println(out)
+
```
PK85, line 50: Create dummy service instance, plan and class names don't matter. With dummy secret it will be in error state.

testing/go/e2e_test.go
```diff
@@ -10,16 +10,54 @@ import (
...
 
-	if len(out) == 0 {
-		t.Error("No CRD found")
+	if !strings.Contains(out, "Running") {
+		t.Error("BTPOperator is not in Running state")
+	}
+
+	// Check that the btp-operator-controller is created and has the expected status.
+	log.Println("kubectl get btp-operator-controller")
+	c, b = exec.Command("kubectl", "get", "btp-operator-controller"), new(strings.Builder)
```
PK85, line 29: We need to add some retries here, e.g for 1 min max.

testing/go/e2e_test.go
PK85: We need to add some retries here, e.g for 1 min max.

PK85 (changes requested):
See my comments

testing/go/e2e_test.go
PK85: Create that yaml file in the testing/go/resources dir

PK85:
@tillknuesting please update this PR with ```Changes proposed in this pull request:```

PR #85, opened 2023-01-04
Fix makefile: module-build to be working

piotrmiskiewicz:
**Description**

- ...
- ...
- ...

**Related issue(s)**


operator/controllers/btpoperator_controller_test.go
```diff
@@ -176,19 +176,6 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
 					secret, err := createSecretWithoutKeys()
 					Expect(err).To(BeNil())
 					Eventually(k8sClient.Create(ctx, secret)).WithTimeout(k8sOpsTimeout).WithPolling(k8sOpsPollingInterval).Should(Succeed())
-					Eventually(getCurrentCrStatus).
```
piotrmiskiewicz, line 179: It is a mistake to expect two states (two assertions). If we expect "error" we should not expect "processing"

PR #91, opened 2023-01-05
Add temporary files to gitignore

piotrmiskiewicz:
**Description**

- add generated files to gitignore

**Related issue(s)**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ bin/
 /hack/operator-secret.yaml
 /charts/btp-manager/
 .DS_Store
+mod/
+template.yaml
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -3,4 +3,6 @@ bin/
 /hack/operator-secret.yaml
 /charts/btp-manager/
 .DS_Store
-.idea/
\ No newline at end of file
+mod/
+template.yaml
```
wozniakjan, line 7: this will exclude `template.yaml` in every directory, do we want to restrict to certain path? you can do that by specifying full relative path to root dir.

wozniakjan (dismissed):

.gitignore
piotrmiskiewicz: you are right

wozniakjan (approved):

wozniakjan (approved):

PR #94, opened 2023-01-09
Make test executions not share resources

wozniakjan:
**Description**

It was identified that sharing resources between tests could significantly contribute to the test flakiness as described in https://github.com/kyma-project/btp-manager/issues/88. This PR tries to remove accidental undesired dependencies between tests. This PR also removes ad-hoc `time.Sleep` lines from the tests and adds extra checks for more deterministic behaviour.

- first commit https://github.com/kyma-project/btp-manager/pull/94/commits/6c751b4ba563458e47a62fb4bb7fb245db8f93f3 - removes this resource reuse between tests
- second commit https://github.com/kyma-project/btp-manager/commit/7f5d016f213a9c870db1a2ad1e7adf606a9c6d8b - creates a copy of btp-operator chart just for tests without webhooks so we don't have to delete them from testenv kube-apiserver as part of the test execution

Issue #88, opened 2023-01-05
Reduce test flakiness


wozniakjan:
**Description**

The tests are really flaky, with a failure rate 43% through the repo history, recently higher.

```
$ gh api repos/kyma-project/btp-manager/actions/workflows | jq '.workflows[] | select(.name == "make-test") | .id'
43814325

$ gh api repos/kyma-project/btp-manager/actions/workflows/43814325/runs?per_page=100 | jq '.total_count'
59

$ gh api repos/kyma-project/btp-manager/actions/workflows/43814325/runs?per_page=100 | jq '.workflow_runs[].conclusion' | awk '/.*/{a=a+1} /failure/{f=f+1} END{print("all:    "a"\nfailed: "f)}'
all:    59
failed: 27
```

The current remediation was to disable frequently failing assertions from tests https://github.com/kyma-project/btp-manager/pull/87, but it's desired to rather spend time refactoring the tests because these assertions ensure functional integrity and are designed to prevent regressions.

**Expected result**

**Actual result**

**Steps to reproduce**

**Troubleshooting**

wozniakjan:
it was identified that there might be two reasons for higher flakiness
* resource state transitions get asserted by polling. However, states can transition faster than the polling period at which point state sampling can miss change and assertion can fail. Remediated by removal of non-critical state assertions https://github.com/kyma-project/btp-manager/pull/87
* resources are shared between tests. The idiomatic approach of starting each test from clean state has not been chosen because the current implementation of deprovisioning is time-consuming. This currently doesn't have any remediation but I would like to propose addressing performance issue in deprovisioning and then reimplementing the tests to not inherit dirty resources from other tests for reuse.

wozniakjan:
here are a couple of places where the execution of the tests is frequently slowed down

```
 27   2023-01-05T14:51:32.826+0100  DEBUG   controller-runtime.test-env starting control plane ...
 28   2023-01-05T14:51:37.470+0100  DEBUG   controller-runtime.test-env installing CRDs ...
...
 44   2023-01-05T14:51:37.736+0100  INFO    Starting workers   ...
 45   2023-01-05T14:51:39.660+0100  INFO    Handling Initial state ...
...
179   2023-01-05T14:51:41.446+0100  INFO    chart consistency check ...
180   2023-01-05T14:51:44.613+0100  INFO    Handling Deleting state ...
[...]

PR #95, opened 2023-01-11
Wrap function calls inside `Eventually()`

wozniakjan:
**Description**

Raw function calls in `Eventually()` blocks need to be wrapped by function literals (a.k.a. lambdas), otherwise functions get called only once and not inside of the `Eventually()` retry logic because only the result of the call is passed as an argument to `Eventually()`.

**Related issue(s)**
depends on https://github.com/kyma-project/btp-manager/pull/94, will require rebase after it merges
related to: https://github.com/kyma-project/btp-manager/issues/88


szwedm (approved):
/lgtm

PR #96, opened 2023-01-11
Prepare release script

piotrmiskiewicz:
**Description**

- prepare a script which makes a release, part 1

**Related issue(s)**

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+# standard bash error handling
+set -o nounset # treat unset variables as an error and exit immediately.
+set -o errexit # exit immediately when a command fails.
+set -E         # needs to be set if we want the ERR trap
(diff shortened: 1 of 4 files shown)


hack/release.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+# standard bash error handling
+set -o nounset # treat unset variables as an error and exit immediately.
+set -o errexit # exit immediately when a command fails.
+set -E         # needs to be set if we want the ERR trap
```
wozniakjan, line 6: consider also `pipefail`
```
set -o pipefail
```
commands joined by pipe `|` can fail but by default, capturing this is set to false.

wozniakjan (approved):
/hold

feel free to unhold if `pipefail` not necessary

wozniakjan (approved):

PR #99, opened 2023-01-13
Replace polling with watch

wozniakjan:
**Description**

Resource state transitions in tests get asserted by polling. However, states can transition faster than the polling period at which point state sampling can miss change and assertion can fail. This PR introduces watch on resource changes to not miss any temporary state transitions.

**Related issue(s)**

depends on: https://github.com/kyma-project/btp-manager/pull/94, https://github.com/kyma-project/btp-manager/pull/95
related to: https://github.com/kyma-project/btp-manager/issues/88

### DIFF
--- a/controllers/btpoperator_controller.go
+++ b/controllers/btpoperator_controller.go
@@ -1047,6 +1047,11 @@ func (r *BtpOperatorReconciler) softDelete(ctx context.Context, gvk schema.Group
 
 	isBinding := gvk.Kind == btpOperatorServiceBinding
 	for _, item := range list.Items {
+		if item.GetDeletionTimestamp().IsZero() {
+			if err := r.Delete(ctx, &item); err != nil {
+				return err
+			}
+		}
 		item.SetFinalizers([]string{})
 		if err := r.Update(ctx, &item); err != nil {
 			return err
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -19,7 +19,6 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"github.com/onsi/ginkgo/v2/types"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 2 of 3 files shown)


controllers/suite_test.go
```diff
@@ -53,15 +63,59 @@ var (
...
+}
+
 func TestAPIs(t *testing.T) {
 
 	RegisterFailHandler(Fail)
 
 	suiteCfg, reporterCfg := GinkgoConfiguration()
 	//Example of usage label:
 	//suiteCfg.LabelFilter = "(test-update)"
+	SetDefaultEventuallyTimeout(time.Second * 5)
```
wozniakjan, line 118: previously, tests configured the same 5 second timeout for each `Eventually()` call, I think it makes sense to set this globally then

controllers/btpoperator_controller.go
```diff
@@ -1047,6 +1047,11 @@ func (r *BtpOperatorReconciler) softDelete(ctx context.Context, gvk schema.Group
 
 	isBinding := gvk.Kind == btpOperatorServiceBinding
 	for _, item := range list.Items {
+		if item.GetDeletionTimestamp().IsZero() {
+			if err := r.Delete(ctx, &item); err != nil {
+				return err
+			}
+		}
```
wozniakjan, line 1054: previously it was only `hardDelete()` that can call deletion on btp-operator resources. However, there is a timeout on `hardDelete()` that can expire before resource deletion is called or can error and never call the deletion, resulting in resources having no `finalizer` but also no `deletionTimestamp` when going through `softDelete()`.

This redundant resource deletion checks if there is no `deletionTimestamp` and ensures it's called at least once for each resource. This improves the robustness of `softDelete` and allows shorter timeout for tests.

szwedm (dismissed):
/lgtm

szwedm (approved):
/lgtm

Issue #172, opened 2023-02-14
Set RBAC roles into resource we need to have access to


PK85:
**Description**

Right now we use wildcards, see: https://github.com/kyma-project/btp-manager/pull/123/files
We must have access only to the resources that we need.

**AC**
- [x] 1) set btp-manager RBAC ac roles to specific resources we need
- [x] 2) automate RBAC check when getting new chart from upstream, [see](https://github.com/kyma-project/btp-manager/blob/main/.github/workflows/auto-update.yml)
- [x] 3) find a way how to managed RBACs when some k8s types are gone from previous BTP Operator release and new BTP Operator release has some new k8s types to be created.
- [x] 4) enhance documentation about auto-update, [here](https://github.com/kyma-project/btp-manager/blob/main/docs/workflows.md#auto-update-chart-and-resources)

wozniakjan:
I missed the docs part

PR #180, opened 2023-02-15
Autogenerate RBAC from btp-operator resources

wozniakjan:
This PR introduces changes to provide a minimal set of RBAC rules for btp-manager automatically configured from the btp-operator chart resources and its integration into the `auto-update` workflow. Capabilities for deleted resources are tracked as delete-only RBAC.

There are also some resources that btp-manager needs access to and they are not part of btp-operator chart (e.g. `Namespaces`), these are handled statically as `kubebuilder:rbac` marker.

closes: https://github.com/kyma-project/btp-manager/issues/172


hack/update/make-module-resources.sh
```diff
@@ -11,11 +11,25 @@ readonly EXISTING_RESOURCES_DELETE_PATH="../../module-resources/delete"
...
 helm template $tag $CHART_PATH --output-dir $HELM_OUTPUT_PATH --values $CHART_OVERRIDES_PATH --namespace "kyma-system"
 
 trap 'rm -rf -- "temp"' EXIT
+
+declare -A rbac
+
+resource() {
+  # for now, lowercase + plural is ok but it might be neccessary to introduce a different
+  # mechanism becausethis rule doesn't apply for all GVK -> resource transformations
+  echo -n "$(yq '.kind' "$1" | tr '[:upper:]' '[:lower:]')s"
```
wozniakjan, line 26: important to highlight this, there are some `kinds` that don't map to `resources` this way, such as `Ingress`. But given there is no `Ingress` in btp-operator chart, it is not important to solve.

wozniakjan: never mind, I decided to provide a map for core resources, CRDs can be added there on demand

ralikio (approved):

PR #191, opened 2023-02-21
Enhance timeouts for unit test on external cluster

### LINKED ISSUE #173: Enhance timeouts for unit test on external cluster
**Description**

Our unit tests fail on an external cluster (e.g. k3d) because of timeout per test (5-second timeout). This task is about providing configuration for timeout per test when the user wants to use an external cluster or debug tests. This document explains how we guide users to adjust testing settings [here](https://github.com/kyma-project/btp-manager/blob/main/docs/testing.md#environment-variables).

**AC**
- [x] 1) Add additional environment variable like `SINGLE_TEST_TIMEOUT` with default value 5sec: [here](https://github.com/kyma-project/btp-manager/blob/main/testing/set-env-vars.sh#L13) with comment that if user plan to debug or run on external cluster they should increase the timeout (for external cluster 15sec should be ok).
- [x] 2) IN out make file for `test` command make `go test ...` timeout, [see](https://github.com/kyma-project/btp-manager/blob/main/Makefile#L87), configurable as environment variable. Set default value to 30sec in the make file. Then add this new environment property [here](https://github.com/kyma-project/btp-manager/blob/main/testing/set-env-vars.sh#L13) too

MarekMichali:
**Description**

Test timeouts cannot be configured in any other way than a change in the source code. This PR allows to configure timeouts using environment variables.

- Add environment variables for test timeouts

**Related issue(s)**

Resolves #173

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,9 @@ OS_ARCH ?= $(shell uname -m)
 # Operating system type
 OS_TYPE ?= $(shell uname)
 
+# Suite timeout must be exported so it can be overwritten by set-env-vars.sh
+export SUITE_TIMEOUT = 30s
+
 # Credentials used for authenticating into the module registry
 # see `kyma alpha mod create --help for more info`
 # MODULE_CREDENTIALS ?= testuser:testpw
@@ -84,7 +87,7 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 
 .PHONY: test
 test: manifests kustomize generate fmt vet envtest ## Run tests.
-	. ./testing/set-env-vars.sh; go test ./... -timeout 30s -coverprofile cover.out -v
+	. ./testing/set-env-vars.sh; go test ./... -timeout $${SUITE_TIMEOUT} -coverprofile cover.out -v
 
 ##@ Build
 
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 
 .PHONY: test
 test: manifests kustomize generate fmt vet envtest ## Run tests.
-	. ./testing/set-env-vars.sh; if [[ -z "$${SUITE_TIMEOUT}" ]]; then go test ./... -timeout ${SUITE_TIMEOUT} -coverprofile cover.out -v; else go test ./... -timeout $${SUITE_TIMEOUT} -coverprofile cover.out -v; fi
+	. ./testing/set-env-vars.sh; go test ./... -timeout $(SUITE_TIMEOUT) -coverprofile cover.out -v
 
 ##@ Build
 
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ OS_ARCH ?= $(shell uname -m)
 OS_TYPE ?= $(shell uname)
 
 # This value is used only if SUITE_TIMEOUT is not exported in set-env-vars.sh and is not specified by the user during the make execution
-SUITE_TIMEOUT ?= 30s
+SUITE_TIMEOUT ?= $${SUITE_TIMEOUT:-30s}
 
 # Credentials used for authenticating into the module registry
 # see `kyma alpha mod create --help for more info`
--- a/Makefile
+++ b/Makefile
@@ -17,8 +17,8 @@ OS_ARCH ?= $(shell uname -m)
 # Operating system type
 OS_TYPE ?= $(shell uname)
 
-# Suite timeout must be exported so it can be overwritten by set-env-vars.sh
-export SUITE_TIMEOUT = 30s
+# This value is used only if SUITE_TIMEOUT is not exported in set-env-vars.sh and is not specified by the user during the make execution
+SUITE_TIMEOUT ?= 30s
 
 # Credentials used for authenticating into the module registry
 # see `kyma alpha mod create --help for more info`
@@ -87,7 +87,7 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 
 .PHONY: test
 test: manifests kustomize generate fmt vet envtest ## Run tests.
-	. ./testing/set-env-vars.sh; go test ./... -timeout $${SUITE_TIMEOUT} -coverprofile cover.out -v
+	. ./testing/set-env-vars.sh; if [[ -z "$${SUITE_TIMEOUT}" ]]; then go test ./... -timeout ${SUITE_TIMEOUT} -coverprofile cover.out -v; else go test ./... -timeout $${SUITE_TIMEOUT} -coverprofile cover.out -v; fi
 
 ##@ Build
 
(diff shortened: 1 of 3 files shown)


testing/set-env-vars.sh
```diff
@@ -11,6 +11,12 @@ KUBEBUILDER_ASSETS="$(${ENVTEST} use ${ENVTEST_K8S_VERSION} --bin-dir ${LOCALBIN
 # not existing cluster by default
 USE_EXISTING_CLUSTER=${USE_EXISTING_CLUSTER:="false"}
 
+# if you plan to debug or run on existing cluster increase the timeout (30 seconds should be ok)
+SINGLE_TEST_TIMEOUT=${SINGLE_TEST_TIMEOUT:="5s"}
+
+# if you plan to debug or run on existing cluster increase the timeout (180 seconds should be ok)
+SUITE_TIMEOUT=30s
```
wozniakjan, line 18:
```suggestion
export SUITE_TIMEOUT=${30s:-SUITE_TIMEOUT}
```
consider having it as a default here but also possible to be easily overwritten from the outside, e.g.
```
SUITE_TIMEOUT=180s make test
```

testing/set-env-vars.sh
```diff
@@ -23,6 +29,8 @@ DISABLE_WEBHOOK_FILTER_FOR_TESTS=${DISABLE_WEBHOOK_FILTER_FOR_TESTS:="false"}
 
 export KUBEBUILDER_ASSETS
 export USE_EXISTING_CLUSTER
+export SINGLE_TEST_TIMEOUT
+export SUITE_TIMEOUT
```
wozniakjan, line 33: nit: I think `export` can be done at the time of variable initialization

Makefile
```diff
@@ -17,6 +17,9 @@ OS_ARCH ?= $(shell uname -m)
 # Operating system type
 OS_TYPE ?= $(shell uname)
 
+# Suite timeout must be exported so it can be overwritten by set-env-vars.sh
+export SUITE_TIMEOUT = 30s
```
wozniakjan, line 21: I think the comment about `export` allowing overwrite is incorrect, export in `Makefile` for make variables is useful for [recursive `Makefiles`](https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html), subshells and subcommands but here it doesn't do much

imho we should stick to [conditional assignment](https://www.gnu.org/software/make/manual/html_node/Conditional-Assignment.html) here like all the other make variables above and allow caller to define this, just default here in case it's not defined at all
```suggestion
SUITE_TIMEOUT ?= 30s
```

controllers/suite_test.go
```diff
@@ -118,7 +118,17 @@ func TestAPIs(t *testing.T) {
 
 	suiteCfg, reporterCfg := GinkgoConfiguration()
 	ReconfigureGinkgo(&reporterCfg, &suiteCfg)
-	SetDefaultEventuallyTimeout(time.Second * 5)
+	singleTestTimeout := os.Getenv("SINGLE_TEST_TIMEOUT")
+	if singleTestTimeout != "" {
+		timeout, err := time.ParseDuration(singleTestTimeout)
+		if err != nil {
+			SetDefaultEventuallyTimeout(time.Second * 5)
```
wozniakjan, line 125: nit: might be worth at least printing the error with some info so user knows tests are running with 5s timeout

Makefile
PK85: @wozniakjan please run a command test. It will not work without export.

testing/set-env-vars.sh
PK85: And this is exactly a case which we have make test command has already SUITE_TIMEOUT set to 30. Agree.

Makefile
MarekMichali: It will not work if `SUITE_TIMEOUT` is not provided in the `set-env-vars` and as I understand we should be able to run test without the script.

testing/set-env-vars.sh
PK85: yes, one-liners will make the file shorter.

PK85 (commented):
I left some comments

testing/set-env-vars.sh
MarekMichali: So should I refactor the whole file to make it consistent?

Makefile
```diff
@@ -84,7 +87,7 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 
 .PHONY: test
 test: manifests kustomize generate fmt vet envtest ## Run tests.
-	. ./testing/set-env-vars.sh; go test ./... -timeout 30s -coverprofile cover.out -v
+	. ./testing/set-env-vars.sh; go test ./... -timeout $${SUITE_TIMEOUT} -coverprofile cover.out -v
```
PK85, line 90: Add a default here, and think about removing that Env variable from the top of make file.

testing/set-env-vars.sh
PK85: yes

controllers/suite_test.go
PK85: In case there is an error, fail it and print error.

Makefile
wozniakjan: ah, I see, I would still suggest at least using [conditional assignment](https://www.gnu.org/software/make/manual/html_node/Conditional-Assignment.html)
```suggestion
export SUITE_TIMEOUT ?= 30s
```

MarekMichali: I will try to find some workaround to get rid of that ugly export.

wozniakjan: it will work ok without the `export` if you use make variable expansion `$()` instead of explicit environment variable expansion `$${}` here
https://github.com/kyma-project/btp-manager/blob/a665e90575e866554186742aff2c29b6752bdb3c/Makefile#L90

https://www.gnu.org/software/make/manual/html_node/Reference.html

Makefile
```diff
@@ -17,6 +17,9 @@ OS_ARCH ?= $(shell uname -m)
 # Operating system type
 OS_TYPE ?= $(shell uname)
 
+# This value is used only if SUITE_TIMEOUT is not exported in set-env-vars.sh and is not specified by the user during the make execution
+SUITE_TIMEOUT ?= 30s
```
wozniakjan, line 21:
```suggestion
SUITE_TIMEOUT ?= $${SUITE_TIMEOUT:-30s}
```

Makefile
```diff
@@ -84,7 +87,7 @@ generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and
 
 .PHONY: test
 test: manifests kustomize generate fmt vet envtest ## Run tests.
-	. ./testing/set-env-vars.sh; go test ./... -timeout 30s -coverprofile cover.out -v
+	. ./testing/set-env-vars.sh; if [[ -z "$${SUITE_TIMEOUT}" ]]; then go test ./... -timeout ${SUITE_TIMEOUT} -coverprofile cover.out -v; else go test ./... -timeout $${SUITE_TIMEOUT} -coverprofile cover.out -v; fi
```
wozniakjan, line 90:
```suggestion
	. ./testing/set-env-vars.sh; go test ./... -timeout $(SUITE_TIMEOUT) -coverprofile cover.out -v
```

wozniakjan (commented):
how about this? no branching, make variable is initialized from potentially existing or not env variable, user can configure in `set-env-vars.sh` as well as explicitly before calling make.

PR #209, opened 2023-03-06
Certs

ukff:


controllers/btpoperator_controller.go
```diff
@@ -1056,14 +1092,32 @@ func (r *BtpOperatorReconciler) watchSecretPredicates() predicate.Funcs {
 			if !ok {
 				return false
 			}
-			return secret.Name == SecretName && secret.Namespace == ChartNamespace
+			if secret.Name == SecretName && secret.Namespace == ChartNamespace {
```
jaroslaw-pieszka, line 1095: How about:

1. extracting this to separate function to avoid code duplication (see line 1111) and
2. using just logical expression:
``` return secret.Namespace == ChartNamespace && (secret.Name == SecretName || secretName == CaSecret || secret.Name == WebhookSecret) ```

File: controllers/btpoperator_controller.go [hunk 2]
```diff
@@ -1073,6 +1127,12 @@ func (r *BtpOperatorReconciler) watchSecretPredicates() predicate.Funcs {
 			if oldSecret.Name == SecretName && oldSecret.Namespace == ChartNamespace {
 				return true
 			}
+			if oldSecret.Name == CaSecret && oldSecret.Namespace == ChartNamespace {
```
jaroslaw-pieszka, line 1130: If you do extraction suggested in previous comment you could just call the same function passing oldSecret as parameter.

controllers/btpoperator_controller.go
```diff
@@ -1112,6 +1172,16 @@ func (r *BtpOperatorReconciler) reconcileConfig(object client.Object) []reconcil
...
 			ReadyCheckInterval, err = time.ParseDuration(v)
+		case "CaCertificateExpiration":
+			CaCertificateExpiration, err = time.ParseDuration(v)
+		case "WebhookCertificateExpiration":
+			WebhookCertificateExpiration, err = time.ParseDuration(v)
+		case "ExpirationBoundary":
+			ExpirationBoundary, err = time.ParseDuration(v)
+		case "rsaKeyBits":
+			var bits int
+			bits, err = strconv.Atoi(v)
```
jaroslaw-pieszka, line 1183: What in case of err != nil?

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return predicateForCertificates(e.ObjectOld)
+		},
+	}
+}
+
+func (r *BtpOperatorReconciler) reconcileCertificates(ctx context.Context, resourcesToApply *[]*unstructured.Unstructured) error {
+	logger := log.FromContext(ctx)
+	logger.Info("checking certifications started")
```
jaroslaw-pieszka, line 1236:
```suggestion
	logger.Info("certificates reconciliation started")
```

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+
+	caSecretExists, err := r.checkIfSecretExists(ctx, CaSecret)
+	if err != nil {
+		return err
+	}
+	if !caSecretExists {
+		logger.Info("ca secret with cert dosent exists.")
```
jaroslaw-pieszka, line 1243:
```suggestion
		logger.Info("ca secret with cert does not exists.")
```

jaroslaw-pieszka: Is `Info` level appropriate? I would go with `Warn` but it is up to you.

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		if err := r.doFullyCertificatesRegenerate(ctx, resourcesToApply); err != nil {
```
jaroslaw-pieszka, line 1244: You use inconsistent prefixes: `fully` and `partial`. The former is an adverb the latter an adjective.
Consider `doFullCertificatesRegeneration` or `regenerateCertificatesFully`

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		return nil
+	}
+	logger.Info("ca secret exists")
+
+	webhookSecretExists, err := r.checkIfSecretExists(ctx, WebhookSecret)
+	if err != nil {
+		return err
+	}
+	if !webhookSecretExists {
+		logger.Info("web secret secret with cert dosent exists.")
```
jaroslaw-pieszka, line 1256:
```suggestion
		logger.Info("web secret secret with cert does not exists.")
```

jaroslaw-pieszka: `Warn`?

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return predicateForCertificates(e.Object)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return predicateForCertificates(e.ObjectOld)
+		},
+	}
+}
+
+func (r *BtpOperatorReconciler) reconcileCertificates(ctx context.Context, resourcesToApply *[]*unstructured.Unstructured) error {
```
jaroslaw-pieszka, line 1234: This is very long function. Consider separating some parts like:
- secret existence validation
- certs validation

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		return err
+	}
+	if !caSecretExists {
+		logger.Info("ca secret with cert dosent exists.")
+		if err := r.doFullyCertificatesRegenerate(ctx, resourcesToApply); err != nil {
+			return err
+		}
+		return nil
+	}
+	logger.Info("ca secret exists")
```
jaroslaw-pieszka, line 1249:
```suggestion
	logger.Info("CA secret exists")
```

jaroslaw-pieszka: CA is the well known abbreviation so I'd rather you used this form in messages at least.

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		return nil
+	}
+	if !signOk {
+		logger.Error(nil, "webhook cert is not signed by correct ca")
+		if err := r.doFullyCertificatesRegenerate(ctx, resourcesToApply); err != nil {
+			return err
+		}
+		return nil
+	}
+	logger.Info("webhook certificate is signed by correct root")
```
jaroslaw-pieszka, line 1287:
```suggestion
	logger.Info("webhook certificate is signed by correct root CA")
```

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+
+	caCertificateValid, err := r.doesCertificateExpireSoon(ctx, CaSecret)
+	if err != nil {
+		logger.Error(err, "ca cert is invalid")
+		return err
+	}
+	if !caCertificateValid {
+		logger.Error(nil, "ca cert is invalid")
```
jaroslaw-pieszka, line 1295: Should we use message telling that it is valid **but expires soon**.

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+	if !signOk {
+		logger.Error(nil, "webhook cert is not signed by correct ca")
+		if err := r.doFullyCertificatesRegenerate(ctx, resourcesToApply); err != nil {
+			return err
+		}
+		return nil
+	}
+	logger.Info("webhook certificate is signed by correct root")
+
+	caCertificateValid, err := r.doesCertificateExpireSoon(ctx, CaSecret)
```
jaroslaw-pieszka, line 1289: The naming is misleading - function checks if certificate expires soon not if it is valid.
Not expired certificate (which is supposed to expire soon) is valid.

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+	if !caCertificateValid {
+		logger.Error(nil, "ca cert is invalid")
+		if err := r.doFullyCertificatesRegenerate(ctx, resourcesToApply); err != nil {
+			return err
+		}
+		return nil
+	}
+	logger.Info("ca certificate is valid")
+
+	webhookCertificateValid, err := r.doesCertificateExpireSoon(ctx, WebhookSecret)
```
jaroslaw-pieszka, line 1303: See previous comment concerning the name of variable.

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+	if err != nil {
+		logger.Error(err, "webhook cert is invalid")
+		return err
+	}
+	if !webhookCertificateValid {
+		logger.Error(nil, "webhook cert is invalid")
```
jaroslaw-pieszka, line 1309: See previous comment concerning message.

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		return false, nil
+	}
+	return true, nil
+}
+
+func (r *BtpOperatorReconciler) doFullyCertificatesRegenerate(ctx context.Context, resourcesToApply *[]*unstructured.Unstructured) error {
+	logger := log.FromContext(ctx)
+	logger.Info("full regeneration started")
+
+	ca, caPk, err := r.generateSelfSignedCertAndAddToApplyList(ctx, resourcesToApply)
```
jaroslaw-pieszka, line 1332: `caPk` is cryptic. I guess it is CA private key.

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+
+func (r *BtpOperatorReconciler) doPartialCertificatesRegenerate(ctx context.Context, resourceToApply *[]*unstructured.Unstructured) error {
+	logger := log.FromContext(ctx)
+	logger.Info("partial regeneration started")
+
+	err := r.generateSignedCertAndAddToApplyList(ctx, resourceToApply, nil, nil)
+	if err != nil {
+		return fmt.Errorf("error while generating signed cert in partial regeneration proccess. %w", err)
+	}
+	logger.Info("partial regeneration success")
```
jaroslaw-pieszka, line 1358:
```suggestion
	logger.Info("partial regeneration succeeded")
```

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+	if err != nil {
+		return []byte{}, nil, fmt.Errorf("while generating self signed cert: %w", err)
+	}
+
+	err = r.appendCertificationDataToUnstructured(resourcesToApply, CaSecret, ca, caPk, CaSecretDataPrefix)
+	if err != nil {
+		return []byte{}, nil, fmt.Errorf("while adding newly generated self signed cert to resoruces to apply: %w", err)
+	}
+
+	logger.Info("generation of self signed cert ok")
```
jaroslaw-pieszka, line 1376:
```suggestion
	logger.Info("generation of self signed cert succeeded")
```

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+
+	webhookCertificate, webhookPrivateKey, err := r.generateSignedCert(ctx, time.Now().Add(WebhookCertificateExpiration), CA, CAPrivateKey)
+	if err != nil {
+		return fmt.Errorf("while generating signed webhookCertificate: %w", err)
+	}
+	err = r.appendCertificationDataToUnstructured(resourcesToApply, WebhookSecret, webhookCertificate, webhookPrivateKey, WebhookSecretDataPrefix)
+	if err != nil {
+		return fmt.Errorf("while adding newly generated signed webhookCertificate to resoruces to apply: %w", err)
+	}
+	logger.Info("generation of signed webhookCertificate ok")
```
jaroslaw-pieszka, line 1392: You introduced kind of convention in logging.
<something> started
<something> ok

but sometimes inconsistent form (verb + noun):
<something> started
<something> success

Please choose one.

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+			return err
+		}
+		ca, ok := secret.Data[r.buildKeyNameWithExtension(CaSecretDataPrefix, CertificatePostfix)]
+		if !ok || ca == nil {
+			return fmt.Errorf("while receiving certificate data from ca secret in reconcilation webhook")
+		}
+		expectedCa = ca
+	}
+
+	for i, u := range *us {
```
jaroslaw-pieszka, line 1460: Is there any meaning of this apart from reference to type i.e. unstructured?

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		default:
+		}
+	}
+
+	return nil
+}
+
+func (r *BtpOperatorReconciler) reconcileCaBundle(u **unstructured.Unstructured, expectedCaBundle []byte) error {
+	if u == nil {
+		return fmt.Errorf("empty u param to reconcileCaBundle")
```
jaroslaw-pieszka, line 1487: 1. `u` is cryptic, refers to type, not meaning
2. message refers to name of function parameters

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		}
+	}
+
+	(*u).Object[WebhooksKey] = webhooks
+	return nil
+}
+
+func (r *BtpOperatorReconciler) reconcileWebhook(ctx context.Context, u **unstructured.Unstructured, expectedCa []byte, kind string) error {
+	if u == nil {
+		return fmt.Errorf("empty u param to reconcileCaBundle")
```
jaroslaw-pieszka, line 1537: as in previous comment

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+	}
+
+	uu := &unstructured.Unstructured{}
```
jaroslaw-pieszka, line 1540: Use meaningful names please.

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+	uu.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "admissionregistration.k8s.io",
+		Version: "v1",
+		Kind:    kind,
+	})
+	err := r.Get(ctx, client.ObjectKey{Name: (*u).GetName(), Namespace: (*u).GetNamespace()}, uu)
+	notFound := k8serrors.IsNotFound(err)
```
jaroslaw-pieszka, line 1547: What if err is nil? What is value of `notFound` then?

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+	ok, err := certs.VerifyIfSecondCertificateIsSignedByFirstCertificate(caCertificate, webhookCertificate)
+	if err != nil {
+		return false, err
+	}
+
+	return ok, nil
+}
+
+func (r *BtpOperatorReconciler) doesCertificateExpireSoon(ctx context.Context, secretName string) (bool, error) {
+	secretData, err := r.getDataFromSecret(ctx, secretName)
```
jaroslaw-pieszka, line 1581: Kind of duplicate of code of `GetCertificateFromSecret`.

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+	}
+	certificateDecoded, _ := pem.Decode(certificate)
+	certificateTemplate, err := x509.ParseCertificate(certificateDecoded.Bytes)
+	if err != nil {
+		return false, err
+	}
+
+	expirationTriggerBound := certificateTemplate.NotAfter.Add(ExpirationBoundary)
+	expiresSoon := time.Now().UTC().After(expirationTriggerBound)
+	return !expiresSoon, nil
```
jaroslaw-pieszka, line 1603: So `doesCertificateExpiresSoon` returns `!expiresSoon`? Weird.

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		return false, err
+	}
+	certificateDecoded, _ := pem.Decode(certificate)
+	certificateTemplate, err := x509.ParseCertificate(certificateDecoded.Bytes)
+	if err != nil {
+		return false, err
+	}
+
+	expirationTriggerBound := certificateTemplate.NotAfter.Add(ExpirationBoundary)
+	expiresSoon := time.Now().UTC().After(expirationTriggerBound)
```
jaroslaw-pieszka, line 1602: `expirationTriggerBound` is the point in time when certificate expires plus boundary (which is negative value).
This approach (negative boundary and `Add` here make it more complex and harder to grasp.
`expiresSoon` is true if it is already expired or expires soon.

controllers/btpoperator_controller.go
jaroslaw-pieszka: And if expires soon we decided not to use it but to regenerate.

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1201,480 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+
+	return buffer.Bytes(), nil
+}
+
+func (r *BtpOperatorReconciler) getValueByKey(key string, data map[string][]byte) ([]byte, error) {
+	value, ok := data[key]
+	if !ok {
+		return nil, fmt.Errorf("while getting data for key: %s", key)
+	}
+	if value == nil || bytes.Equal(value, []byte{}) {
```
jaroslaw-pieszka, line 1661:
```suggestion
	if value == nil || len(value) == 0 {
```

controllers/btpoperator_controller_test.go
```diff
@@ -872,3 +1194,97 @@ func checkIfNoBtpResourceExists() {
 func canIgnoreErr(err error) bool {
 	return k8serrors.IsNotFound(err) || meta.IsNoMatchError(err) || k8serrors.IsMethodNotSupported(err)
 }
+
+func replaceCaBundleInValidatingWebhooks(i []byte) bool {
```
jaroslaw-pieszka, line 1198: `i` is a CA bundle? Name is misleading.

controllers/btpoperator_controller_test.go
```diff
@@ -872,3 +1194,97 @@ func checkIfNoBtpResourceExists() {
...
+
+func ensureAllWebhooksManagedByBtpOperatorHaveCorrectCABundles() {
+	secret := getSecret(CaSecret)
+	ca, ok := secret.Data[reconciler.buildKeyNameWithExtension(CaSecretDataPrefix, CertificatePostfix)]
+	Expect(ok).To(BeTrue())
+	Expect(ca).To(Not(BeNil()))
+	vw := &admissionregistrationv1.ValidatingWebhookConfigurationList{}
+	err := k8sClient.List(ctx, vw, managedByLabelFilter)
+	Expect(err).To(BeNil())
+	for _, w := range vw.Items {
```
jaroslaw-pieszka, line 1277: Loops could be extracted as a separate function and used for both kinds of webhooks.

internal/certs/certs.go
```diff
@@ -0,0 +1,156 @@
...
+
+func VerifyIfSecondCertificateIsSignedByFirstCertificate(firstCertificate, secondCertificate []byte) (bool, error) {
+	firstCertificateDecoded, _ := pem.Decode(firstCertificate)
+	secondCertificateDecoded, _ := pem.Decode(secondCertificate)
+
+	firstCertificateTemplate, err := x509.ParseCertificate(firstCertificateDecoded.Bytes)
+	if err != nil {
+		return false, err
+	}
+	if !firstCertificateTemplate.IsCA {
```
jaroslaw-pieszka, line 123: Should we check also if it is self-signed CA root? Or just checking if it is CA cert is enough?

internal/certs/certs.go
```diff
@@ -0,0 +1,156 @@
...
+		return false, err
+	}
+
+	if secondCertificateTemplate.IsCA {
+		return false, fmt.Errorf("certificate given as second is CA")
+	}
+
+	_, err = secondCertificateTemplate.Verify(verifyOpts)
+	if err != nil {
+		return false, fmt.Errorf("verify of second certificate from first certificate error: %w", err)
```
jaroslaw-pieszka, line 148: Kind of vague message (referring to order of parameters which).

internal/certs/certs.go
```diff
@@ -0,0 +1,156 @@
...
+	newCertificatePrivateKeyPem := new(bytes.Buffer)
+	pem.Encode(newCertificatePrivateKeyPem, &pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(newCertificatePrivateKey),
+	})
+
+	return newCertificatePem.Bytes(), newCertificatePrivateKeyPem.Bytes(), nil
+}
+
+func VerifyIfSecondCertificateIsSignedByFirstCertificate(firstCertificate, secondCertificate []byte) (bool, error) {
```
jaroslaw-pieszka, line 115: Consider other naming e.g. caCertificate, leafCertificate
VerifyIfLeafSignedByGivenCA or so.

controllers/btpoperator_controller.go
jaroslaw-pieszka: The meaning from my perspective is either:
- cert is expired or will expire soon
- we should regenerate it

controllers/btpoperator_controller_test.go
jaroslaw-pieszka: I see it now. Thanks.

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1178,506 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		return err
+	}
+	if breakFlow {
+		return nil
+	}
+
+	return nil
+}
+
+func (r *BtpOperatorReconciler) checkIfCertificatesExists(ctx context.Context, resourcesToApply *[]*unstructured.Unstructured) (bool, error) {
```
jaroslaw-pieszka, line 1247: `checkIfCertificatesExist` would be better choice.

preparemain.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
jaroslaw-pieszka, line 1: Is it added on purpose? Or accidentally?

main.go
```diff
@@ -52,6 +53,10 @@ func init() {
 }
 
 func main() {
+	cmd := exec.Command("/bin/sh", "preparemain.sh")
```
jaroslaw-pieszka, line 56: Looks like some temporary testing code.

afterrunmain.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
jaroslaw-pieszka, line 1: Is it added on purpose? Or accidentally?

docs/certs.md
```diff
@@ -0,0 +1,10 @@
+![Certification management diagram](./assets/certs.svg)
+
+## Certification management
+
+BTP Manager maintain two secrets `ca-server-cert` and `webhook-server-cert` which are used to allow communication within BTP Operator webhooks, and in result allow to create resources like ServiceInstances and ServiceBindings
```
IwonaLanger, line 5:
```suggestion
BTP Manager maintains two secrets, `ca-server-cert` and `webhook-server-cert`, which are used to allow communication within BTP Operator webhooks and thus allow the creation of resources like ServiceInstances and ServiceBindings.
```

docs/certs.md
```diff
@@ -0,0 +1,10 @@
...
+During provisioning, first there is created `ca-server-cert` which is self sign CA certificate, then based on that, application create signed cert `webhook-server-cert`, which is mounted under the deployment.
```
IwonaLanger, line 6:
```suggestion
During provisioning, at first, `ca-server-cert` is created. It is a self-sign CA certificate. Then based on that, the application creates a signed cert, `webhook-server-cert`, which is mounted under the deployment.
```

docs/certs.md
```diff
@@ -0,0 +1,10 @@
...
+Webhooks have CABundle field which is set to the content of ca-server-cert and this field is managed by BTP Manager.
```
IwonaLanger, line 7:
```suggestion
Webhooks have a CABundle field set to the content of `ca-server-cert,` and BTP Manager manages this field.
```

IwonaLanger, line 7: You use `caBundle`, `CA bundle`, `ca bundle` and `CABundle`. If they're not three different things, why don't you use the same spelling in the whole doc?

docs/certs.md
```diff
@@ -0,0 +1,10 @@
...
+The `ca-server-cert`, `webhook-server-cert`, and webhooks caBundles are keep in sync by using reconcilation mechanism, which mean every manual change by user in this resources will trigger automatically regeneration of all three resources.
```
IwonaLanger, line 8:
```suggestion
The `ca-server-cert`, `webhook-server-cert`, and webhooks' caBundles are kept in sync by using the reconciliation mechanism, which means every manual change in these resources by the user will automatically trigger regeneration of all three resources.
```

docs/certs.md
```diff
@@ -0,0 +1,10 @@
...
+BTP Manager maintain resources by creating, deleting and updating actions during reconcilation. The goal is to keep `ca-server-cert`, `webhook-servert-cert` and webhooks ca bundles in sync, all the time.
```
IwonaLanger, line 9:
```suggestion
BTP Manager maintains resources by creating, deleting, and updating actions during reconciliation. The goal is to keep `ca-server-cert`, `webhook-server-cert`, and webhooks' ca bundles in sync all the time.
```

docs/certs.md
```diff
@@ -0,0 +1,10 @@
...
+Scheduled reconcilation check also expiration dates of certificates and if it detected that certificate expires soon, it regenerates it in advanced, to keep things smoothly.
```
IwonaLanger, line 10:
```suggestion
Scheduled reconciliation also checks expiration dates of certificates, and if it detects that a certificate expires soon, it regenerates it in advance so that the processes run smoothly.
```

internal/certs/certs.go
```diff
@@ -0,0 +1,175 @@
...
+		return []byte{}, nil, err
+	}
+
+	newCertificate, err := x509.CreateCertificate(rand.Reader, newCertificateTemplate, newCertificateTemplate, &newCertificatePrivateKey.PublicKey, newCertificatePrivateKey)
+	if err != nil {
+		return []byte{}, nil, err
+	}
+
+	newCertificatePem := new(bytes.Buffer)
+	pem.Encode(newCertificatePem, &pem.Block{
```
szwedm, line 55: ignored err

internal/certs/certs.go
```diff
@@ -0,0 +1,175 @@
...
+		Type:  "CERTIFICATE",
+		Bytes: newCertificate,
+	})
+
+	newCertificatePrivateKeyPem := new(bytes.Buffer)
+	pem.Encode(newCertificatePrivateKeyPem, &pem.Block{
```
szwedm, line 61: ignored err

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+	if breakFlow {
+		return nil
+	}
+
+	return nil
+}
+
+func (r *BtpOperatorReconciler) checkIfCertificatesExist(ctx context.Context, resourcesToApply *[]*unstructured.Unstructured) (bool, error) {
+	logger := log.FromContext(ctx)
+	caSecretExists, err := r.checkIfSecretExists(ctx, CaSecret)
```
szwedm, line 1343: `checkIfSecretExists` doesn't return error even if it occurred

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+
+	unstructuredObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(secret)
+	if err != nil {
+		return err
+	}
+	*us = append(*us, &unstructured.Unstructured{Object: unstructuredObj})
+	return nil
+}
+
+func (r *BtpOperatorReconciler) mapCertToSecretData(certificate, privateKey []byte, keyNameForCert, keyNameForPrivateKey string) (error, map[string][]byte) {
```
szwedm, line 1546: why is error in return values?

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+	}
+	if value == nil || len(value) == 0 {
+		return nil, fmt.Errorf("empty data for key: %s", key)
+	}
+	return value, nil
+}
+
+func (r *BtpOperatorReconciler) buildSecretWithDataAndLabels(name string, data map[string][]byte, labels map[string]string) *corev1.Secret {
+	return &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{
```
szwedm, line 1734: Kind and APIVersion should be populated by default when you create a Secret, no need to specify TypeMeta

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+
+	webhookCertificate, webhookPrivateKey, err := certs.GenerateSignedCertificate(expiration, caCertificate, caPrivateKey)
+	if err != nil {
+		return []byte{}, nil, err
+	}
+
+	return webhookCertificate, webhookPrivateKey, err
+}
+
+func (r *BtpOperatorReconciler) appendCertificationDataToUnstructured(us *[]*unstructured.Unstructured, certName string, certificate, privateKey []byte, prefix string) error {
```
szwedm, line 1531: mixed order of params in the signature of method

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+	return nil
+}
+
+func (r *BtpOperatorReconciler) generateSelfSignedCertAndAddToApplyList(ctx context.Context, resourcesToApply *[]*unstructured.Unstructured) ([]byte, []byte, error) {
+	logger := log.FromContext(ctx)
+	logger.Info("generation of self signed cert started")
+
+	caCertificate, caPrivateKey, err := certs.GenerateSelfSignedCertificate(time.Now().Add(CaCertificateExpiration))
+	if err != nil {
+		return []byte{}, nil, fmt.Errorf("while generating self signed cert: %w", err)
```
szwedm, line 1477: why `[]byte{}` is returned in case of error occurrence?

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+	}
+	logger.Info("partial regeneration succeeded")
+	return nil
+}
+
+func (r *BtpOperatorReconciler) generateSelfSignedCertAndAddToApplyList(ctx context.Context, resourcesToApply *[]*unstructured.Unstructured) ([]byte, []byte, error) {
+	logger := log.FromContext(ctx)
+	logger.Info("generation of self signed cert started")
+
+	caCertificate, caPrivateKey, err := certs.GenerateSelfSignedCertificate(time.Now().Add(CaCertificateExpiration))
```
szwedm, line 1475: typo in `caCertificate`

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+	}
+	logger.Info("generation of signed webhook certificate success")
+	return nil
+}
+
+func (r *BtpOperatorReconciler) generateSignedCert(ctx context.Context, expiration time.Time, caCertificate, caPrivateKey []byte) ([]byte, []byte, error) {
+	if caCertificate == nil || caPrivateKey == nil {
+		data, err := r.getDataFromSecret(ctx, CaSecret)
+		if err != nil {
+			return []byte{}, nil, err
```
szwedm, line 1509: why `[]byte{}` is returned instead of nil?

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		}
+
+		caCertificate, err = r.getValueByKey(r.buildKeyNameWithExtension(CaSecretDataPrefix, CertificatePostfix), data)
+		if err != nil {
+			return []byte{}, nil, err
```
szwedm, line 1514: why `[]byte{}` is returned instead of nil?

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		}
+
+		caPrivateKey, err = r.getValueByKey(r.buildKeyNameWithExtension(CaSecretDataPrefix, RsaKeyPostfix), data)
+		if err != nil {
+			return []byte{}, nil, err
```
szwedm, line 1519: why `[]byte{}` is returned instead of nil?

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		}
+	}
+
+	webhookCertificate, webhookPrivateKey, err := certs.GenerateSignedCertificate(expiration, caCertificate, caPrivateKey)
+	if err != nil {
+		return []byte{}, nil, err
```
szwedm, line 1525: why `[]byte{}` is returned instead of nil?

internal/certs/certs.go
```diff
@@ -0,0 +1,175 @@
...
+		DNSNames:     getDns(),
+		NotBefore:    time.Now(),
+		NotAfter:     expiration,
+		ExtKeyUsage:  []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+		KeyUsage:     x509.KeyUsageDigitalSignature,
+	}
+
+	newCertificatePrivateKey, err := rsa.GenerateKey(rand.Reader, RsaKeyBits())
+	if err != nil {
+		return []byte{}, nil, err
```
szwedm, line 81: why return `[]byte{}` here and in the subsequent returns?

internal/certs/certs.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	newCertificate, err := x509.CreateCertificate(rand.Reader, newCertificateTemplate, parsedSourceCertificate, &newCertificatePrivateKey.PublicKey, parsedSourcePrivateKey)
+	if err != nil {
+		return []byte{}, nil, err
+	}
+
+	newCertificatePem := new(bytes.Buffer)
+	pem.Encode(newCertificatePem, &pem.Block{
```
szwedm, line 107: ignored err

internal/certs/certs.go
```diff
@@ -0,0 +1,175 @@
...
+		Type:  "CERTIFICATE",
+		Bytes: newCertificate,
+	})
+
+	newCertificatePrivateKeyPem := new(bytes.Buffer)
+	pem.Encode(newCertificatePrivateKeyPem, &pem.Block{
```
szwedm, line 113: ignored err

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+func (r *BtpOperatorReconciler) mapCertToSecretData(certificate, privateKey []byte, keyNameForCert, keyNameForPrivateKey string) (error, map[string][]byte) {
+	return nil, map[string][]byte{
+		keyNameForCert:       certificate,
+		keyNameForPrivateKey: privateKey,
+	}
+}
+
+func (r *BtpOperatorReconciler) reconcileWebhooksConfigurations(ctx context.Context, resourcesToApply *[]*unstructured.Unstructured, expectedCa []byte) error {
+	logger := log.FromContext(ctx)
+	logger.Info("starting reconcilation of webhooks")
```
szwedm, line 1555: Typo: In word 'reconcilation'

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		}
+		ca, ok := secret.Data[r.buildKeyNameWithExtension(CaSecretDataPrefix, CertificatePostfix)]
+		if !ok || ca == nil {
+			return fmt.Errorf("while receiving certificate data from CA secret in reconcilation webhook")
+		}
+		expectedCa = ca
+	}
+
+	for i, resource := range *resourcesToApply {
+		switch kind := resource.GetKind(); kind {
```
szwedm, line 1570: I think if instead of switch would be more readable since there is only one case

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+		}
+	}
+
+	logger.Info("webhooks cert bundles check success")
+	return nil
+}
+
+func (r *BtpOperatorReconciler) reconcileWebhook(ctx context.Context, webhook **unstructured.Unstructured, expectedCa []byte) error {
+	logger := log.FromContext(ctx)
+	if webhook == nil {
```
szwedm, line 1589: is it possible that `webhook` will be nil? The method receives args from for...range loop through `*unstructured.Unstructured`

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+
+	for i, resource := range *resourcesToApply {
+		switch kind := resource.GetKind(); kind {
+		case MutatingWebhookConfiguration, ValidatingWebhookConfiguration:
+			{
+				err := r.reconcileWebhook(ctx, &resource, expectedCa)
+				if err != nil {
+					return err
+				}
+				(*resourcesToApply)[i] = resource
```
szwedm, line 1577: `reconcileWebhook` mutates the webook pointed by `resource` var, the changes should be already reflected in the slice without the need to overwrite the element in the slice

controllers/btpoperator_controller.go
```diff
@@ -1131,3 +1179,567 @@ func (r *BtpOperatorReconciler) watchConfigPredicates() predicate.Funcs {
...
+}
+
+func (r *BtpOperatorReconciler) mapCertToSecretData(certificate, privateKey []byte, keyNameForCert, keyNameForPrivateKey string) (error, map[string][]byte) {
+	return nil, map[string][]byte{
+		keyNameForCert:       certificate,
+		keyNameForPrivateKey: privateKey,
+	}
+}
+
+func (r *BtpOperatorReconciler) reconcileWebhooksConfigurations(ctx context.Context, resourcesToApply *[]*unstructured.Unstructured, expectedCa []byte) error {
```
szwedm, line 1553: think about changing the `*[]*unstructured.Unstructured` param to just `[]*unstructured.Unstructured` when there is no need to change the length of the slice. Imo, it could improve the readability of the code

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+		orgWebhookCertExpiration := WebhookCertificateExpiration
+		orgExpirationBoundary := ExpirationBoundary
+
+		var actualWorkqueueSize func() int
+
+		BeforeAll(func() {
+			Expect(removeAllFromPath(defaultChartPath)).To(Succeed())
+			Expect(removeAllFromPath(defaultResourcesPath)).To(Succeed())
+
+			os.Setenv("DISABLE_WEBHOOK_FILTER_FOR_TESTS", "true")
```
szwedm, line 531: unhandled error

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
 			})
 		})
 	})
+
+	Describe("Certification management", func() {
+		orgCaCertificateExpiration := CaCertificateExpiration
+		orgWebhookCertExpiration := WebhookCertificateExpiration
+		orgExpirationBoundary := ExpirationBoundary
+
+		var actualWorkqueueSize func() int
```
szwedm, line 525: I suggest to combine var declaration in L525 with var definition in L537

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+			os.Setenv("DISABLE_WEBHOOK_FILTER_FOR_TESTS", "true")
+			Expect(createChartOrResourcesCopyWithoutWebhooks("../module-chart/chart", defaultChartPath)).To(Succeed())
+			Expect(createChartOrResourcesCopyWithoutWebhooks("../module-resources", defaultResourcesPath)).To(Succeed())
+			ChartPath = defaultChartPath
+			ResourcesPath = defaultResourcesPath
+
+			actualWorkqueueSize = func() int { return reconciler.workqueueSize }
+		})
+
+		restoreOriginalCertificateTiems := func() {
```
szwedm, line 540: typo in `restoreOriginalCertificateTiems`

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+			cr = &v1alpha1.BtpOperator{}
+			Expect(k8sClient.Get(ctx, client.ObjectKey{Namespace: defaultNamespace, Name: btpOperatorName}, cr)).Should(Succeed())
+			Expect(k8sClient.Delete(ctx, cr)).Should(Succeed())
+			Eventually(updateCh).Should(Receive(matchState(types.StateReady)))
+			Eventually(updateCh).Should(Receive(matchDeleted()))
+			Expect(isCrNotFound()).To(BeTrue())
+			restoreOriginalCertificateTiems()
+		}
+
+		ensureReconcilationQueueIsEmpty := func() {
```
szwedm, line 571: typo in `ensureReconcilationQueueIsEmpty `

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+			Eventually(actualWorkqueueSize).WithTimeout(time.Second * 5).WithPolling(time.Millisecond * 100).Should(Equal(0))
```
szwedm, line 572: I'm not sure whether `Eventually` won't poll returned value from `actualWorkqueueSize` instead of running the func every time. Maybe it would be better to pass a lambda to `Eventually` instead of var holding a func

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+		ensureCorrectState := func() {
+			ensureReconcilationQueueIsEmpty()
+			ok, err := reconciler.isWebhookSecretCertSignedByCaSecretCert(ctx)
+			Expect(err).To(BeNil())
+			Expect(ok).To(BeTrue())
+			ensureAllWebhooksManagedByBtpOperatorHaveCorrectCABundles()
+		}
+
+		When("doing provisioning", func() {
+			BeforeAll(func() {
```
szwedm, line 584: common setup node for some specs, I suggest to extract the common setup to `BeforeEach` setup node and place it a new container node like `Describe` or `Context`

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+				certBeforeEach(nil)
+			})
+			AfterAll(func() {
```
szwedm, line 587: common setup node for some specs, I suggest to extract the common setup to `AfterEach` setup node and place it a new container node like `Describe` or `Context`

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+
+		ensureCorrectState := func() {
+			ensureReconcilationQueueIsEmpty()
+			ok, err := reconciler.isWebhookSecretCertSignedByCaSecretCert(ctx)
+			Expect(err).To(BeNil())
+			Expect(ok).To(BeTrue())
+			ensureAllWebhooksManagedByBtpOperatorHaveCorrectCABundles()
+		}
+
+		When("doing provisioning", func() {
```
szwedm, line 583: I think `doing provisioning` is too broad to describe the context of test. How about changing the description to `certs don't exist in the cluster prior provisioning` or `provisioning the module for the first time`

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+			AfterAll(func() {
+				certAfterEach()
+			})
+			It("should do fully regenerate of CA certificate and webhook certificate", func() {
+				newCaCertificate, newCaPrivateKey, err := certs.GenerateSelfSignedCertificate(time.Now().Add(CaCertificateExpiration))
+				newCaPrivateKeyStructured, err := structToByteArray(newCaPrivateKey)
+				Expect(err).To(BeNil())
+
+				caSecret := getSecret(CaSecret)
+				orgCaSecret := caSecret
```
szwedm, line 608: both vars point to the same Secret

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+				Expect(err).To(BeNil())
+				currentWebhookSecret := getSecret(WebhookSecret)
+				originalWebhookSecret := currentWebhookSecret
+
+				newWebhookCertificate, newWebhookPrivateKey, err := certs.GenerateSignedCertificate(time.Now().Add(WebhookCertificateExpiration), ca, pk)
+				Expect(err).To(BeNil())
+				newWebhookPrivateKeyStructured, err := structToByteArray(newWebhookPrivateKey)
+				Expect(err).To(BeNil())
+
+				webhookCert := getSecret(CaSecret)
```
szwedm, line 657: looks like getting wrong Secret, shouldn't it be `WebhookSecret`?

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+
+				afterCaSecret := getSecret(CaSecret)
+				afterCaSecretCert, ok := afterCaSecret.Data[reconciler.buildKeyNameWithExtension(CaSecretDataPrefix, CertificatePostfix)]
+				beforeCaSecretCert, ok := beforeCaSecret.Data[reconciler.buildKeyNameWithExtension(CaSecretDataPrefix, CertificatePostfix)]
+				Expect(bytes.Equal(afterCaSecretCert, beforeCaSecretCert))
+				ensureCorrectState()
+			})
+		})
+
+		When("webhook certificate and is signed by different CA certificate", func() {
```
szwedm, line 677: typo in container node description `webhook certificate and is...`

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+				Expect(ok).To(BeTrue())
+				Expect(!bytes.Equal(currentWebhookCert, beforeWebhookCert))
+				Expect(!bytes.Equal(currentWebhookCert, newWebhookCertificate))
+
+				ensureCorrectState()
+			})
+		})
+
+		When("webhook certificate expires", func() {
+			fakeTime := 30.0
```
szwedm, line 719: you can set time.Duration instead of float

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+				certBeforeEach(timeOpts)
+			})
+			AfterAll(func() {
+				certAfterEach()
+			})
+
+			It("CA certificate is not changed, webhook certificate is regenerated", func() {
+				caSecretBeforeExpiration := getSecret(CaSecret)
+				webhookSecretBeforeExpiration := getSecret(WebhookSecret)
+				Expect(checkHowManySecondsToExpiration(WebhookSecret) <= fakeTime).To(BeTrue())
```
szwedm, line 736: you can use gomega numerical matcher: https://onsi.github.io/gomega/#working-with-numbers-and-times

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+			})
+			AfterAll(func() {
+				certAfterEach()
+			})
+
+			It("fully regenerate of CA certificate and webhook certificate", func() {
+				caSecretBeforeExpiration := getSecret(CaSecret)
+				webhookSecretBeforeExpiration := getSecret(WebhookSecret)
+				Expect(checkHowManySecondsToExpiration(CaSecret) <= fakeSeconds).To(BeTrue())
+				time.Sleep(time.Second * 10)
```
szwedm, line 774: you can poll `time.After` to check whether cert expired instead of using `time.Sleep`

controllers/btpoperator_controller_test.go
```diff
@@ -919,3 +1242,108 @@ func checkIfNoBtpResourceExists() {
...
+	err := k8sClient.List(ctx, webhookConfig, managedByLabelFilter)
+	Expect(err).To(BeNil())
+	Expect(len(webhookConfig.Items) > 0).To(BeTrue())
+	if len(webhookConfig.Items) > 0 {
+		webhook := webhookConfig.Items[0]
+		if len(webhook.Webhooks) > 0 {
+			webhook.Webhooks[0].ClientConfig.CABundle = newCaBundle
+			err := k8sClient.Update(ctx, &webhook)
+			Expect(err).To(BeNil())
+			//time.Sleep(time.Second * 10)
```
szwedm, line 1275: remove unused code

controllers/btpoperator_controller_test.go
```diff
@@ -500,6 +516,313 @@ var _ = Describe("BTP Operator controller", Ordered, func() {
...
+		When("webhook caBundle modified with some dummy text", func() {
+			BeforeAll(func() {
+				certBeforeEach(nil)
+			})
+			AfterAll(func() {
+				certAfterEach()
+			})
+			It("should be reconciled to existing CA certificate", func() {
+				dummy := []byte("dummy")
+				updated := replaceCaBundleInMutatingWebhooks(dummy)
```
szwedm, line 820: no assertion for `updated` var

testing/set-env-vars.sh
```diff
@@ -15,7 +15,7 @@ export USE_EXISTING_CLUSTER=${USE_EXISTING_CLUSTER:="false"}
 export SINGLE_TEST_TIMEOUT=${SINGLE_TEST_TIMEOUT:="5s"}
 
 # if you plan to debug or run on existing cluster increase the timeout (180 seconds should be ok)
-export SUITE_TIMEOUT=${SUITE_TIMEOUT:=30s}
+export SUITE_TIMEOUT=${SUITE_TIMEOUT:=40s}
```
szwedm, line 18: can you explain why do you extend timeout for test suite execution?

docs/assets/.$certs.drawio.bkp
szwedm, line 1: why is this file required?

controllers/btpoperator_controller.go
```diff
@@ -1212,3 +1326,518 @@ func (r *BtpOperatorReconciler) IsForceDelete(cr *v1alpha1.BtpOperator) bool {
...
+	}
+	logger.Info("webhook secret exists")
+	return false, nil
+}
+
+func (r *BtpOperatorReconciler) ensureSecretsDataIsSet(ctx context.Context, resourcesToApply *[]*unstructured.Unstructured) (bool, error) {
+	caSecretData, err := r.getDataFromSecret(ctx, CaSecret)
+	caSecretDataIncorrect := false
+	_, err = r.getValueByKey(r.buildKeyNameWithExtension(CaSecretDataPrefix, CertificatePostfix), caSecretData)
+	if err != nil {
```
jaroslaw-pieszka, line 1416: `caSecretDataIncorrect := err != nil` would do. (3 lines less)

docs/certs.md
```diff
@@ -0,0 +1,12 @@
+---
+Certification management
+---
+
+![Certification management diagram](./assets/certs.svg)
+
```
IwonaLanger, line 6: Some more comments on the diagram:
1. The font color in the third diamond from the left seems inconsistent with the rest of them.
2. The questions in the last block (Expiration) are incorrect. Remove `does` from the last two diamonds. How about rephrasing them to `cert from (...) secret expires within a given time?`?
3. Remove `is` from the fourth diamond and `no` from the first two diamonds.
4. You use `CA bundle` in the diagram and `caBundle` in the description. Shouldn't they be the same?
5. In the top left-hand corner of the diagram there's `self signed`. Change it to `self-signed`.
6. If there are questions, there should be some answers. How about adding yes/no arrows?

controllers/btpoperator_controller.go
```diff
@@ -1124,29 +1151,104 @@ func (r *BtpOperatorReconciler) enqueueOldestBtpOperator() []reconcile.Request {
...
 			if !ok {
 				return false
 			}
-			if oldSecret.Name == SecretName && oldSecret.Namespace == ChartNamespace {
+			return predicateIfReconcile(oldSecret)
+		},
+	}
+}
+
+func (r *BtpOperatorReconciler) watchValidatingWebhooksPredicates() predicate.Funcs {
```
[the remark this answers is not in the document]
jaroslaw-pieszka: I beg to disagree. But it is not critical, we should proceed with this PR. I will create a small PR afterwards to show you what I meant.

controllers/btpoperator_controller.go
jaroslaw-pieszka: At the beginning `caSecretDataIncorrect` is `false`.
The logic states that if err != nil `caSecretDataIncorrect` is set to `true`. Otherwise it is still `false`.
So this is equivalent of `caSecretDataIncorrect := err != nil`.
Nothing is overwritten.

In subsequent `ifs` you can use:
`caSecretDataIncorrect = caSecretDataIncorrect || err != nil`

controllers/btpoperator_controller.go
```diff
@@ -417,11 +434,11 @@ func (r *BtpOperatorReconciler) getResourcesToApplyPath() string {
 	return fmt.Sprintf("%s%capply", ResourcesPath, os.PathSeparator)
 }
 
-func (r *BtpOperatorReconciler) prepareModuleResources(ctx context.Context, us []*unstructured.Unstructured, s *corev1.Secret) error {
+func (r *BtpOperatorReconciler) prepareModuleResourcesFromManifests(ctx context.Context, resourcesToApply []*unstructured.Unstructured, s *corev1.Secret) error {
 	logger := log.FromContext(ctx)
 
 	var configMapIndex, secretIndex int
-	for i, u := range us {
+	for i, u := range resourcesToApply {
```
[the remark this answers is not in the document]
jaroslaw-pieszka: Is is possible that we won't find secret in the resources? How we should handle this?
The same for configMap. We should decide whether we should handle or ignore (leave as it is).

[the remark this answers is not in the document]
szwedm: Both, Secret and ConfigMap, are part of the module resources atm, but we can't be sure whether they are here to stay. We should handle the case when they are missing, throw an error and fail fast since they are required for the module to work.

szwedm:
All my remarks have been addressed.
/lgtm

controllers/btpoperator_controller.go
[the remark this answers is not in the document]
szwedm: I will create a separate PR with fix

IwonaLanger (approved):

szwedm (approved):

PR #229, opened 2023-03-14
Set harddelete status when force-delete label is added during deletion

piotrmiskiewicz:
**Description**

- if force-delete label is added when deletion is blocked - we expect also HardDelete reason in the status

**Related issue(s)**


szwedm (dismissed):
/lgtm

szwedm (approved):
/lgtm

Issue #228, opened 2023-03-14
React on SI/SB removals and reconcile BTP Operator CR


PK85:
**AC**:
- when the user removes SI/SB then immediately reconcile BTP Operator CR

PK85:
See:
- https://github.com/kubernetes-sigs/controller-runtime/issues/540
- https://github.com/kubernetes-sigs/controller-runtime/pull/2159

szwedm:
@piotrmiskiewicz we can decrease reconciliation interval from 15 mins to 3 mins

piotrmiskiewicz:
PR: https://github.com/kyma-project/btp-manager/pull/240

PR #240, opened 2023-03-22
React on sisb deletion

piotrmiskiewicz:
**Description**

- the main controller starts/stops ServiceInstance controller which watches if the last SI (SB) was removed. If yes, and the BTP operator is in deletion state (and watining for SI/SB removal) then triggers btpoperator reconciliation by changing state to HardDelete
- small changes in env tests

**Related issue(s)**
https://github.com/kyma-project/btp-manager/issues/228


piotrmiskiewicz:
testing scenario:
1. run k3d, install latest btp manager, for example: `/run_module_image.sh europe-docker.pkg.dev/kyma-project/prod/unsigned/component-descriptors/kyma.project.io/module/btp-operator:v0.3.3`
2. modify the image - use from this PR
3. Apply secret and operator yamls: `kubectl apply -f examples/btp-manager-secret.yaml` and `kubectl apply -f examples/btp-operator.yaml`
4. create service instance, for example:
```yaml
apiVersion: services.cloud.sap.com/v1alpha1
kind: ServiceInstance
metadata:
  name: btp-auditlog-4
  namespace: default
spec:
  serviceOfferingName: auditlog-api
  servicePlanName: default
  externalName: btp-auditlog-2
```
5. Delete btp operator resource (deletion is stopped)
6. delete service instance (and remove finalizer from the SI)
7. expected result: btp operator is finally removed

controllers/instnace_binding_controller_manager.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (r *InstanceBindingControllerManager) DisableSISBController() {
+	r.sisbControllerMu.Lock()
+	defer r.sisbControllerMu.Unlock()
+	if !r.enabled {
+		return
+	}
+
+	r.enabled = false
```
szwedm, line 80: I think we should set `r.enabled` to `false` after we are sure that controller has been stopped. Do we have to call `r.stopper()` in a separate goroutine? From go docs:
```
A CancelFunc tells an operation to abandon its work. A CancelFunc does not wait for the work to stop. A CancelFunc may be called by multiple goroutines simultaneously. After the first call, subsequent calls to a CancelFunc do nothing.
```

controllers/instnace_binding_controller_manager.go
```diff
@@ -0,0 +1,87 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sync"
+)
+
+// InstanceBindingControllerManager runs and stops the ServiceInstance controller
+type InstanceBindingControllerManager struct {
+	client.Client
+	*rest.Config
+	Scheme        *runtime.Scheme
+	workqueueSize int
```
szwedm, line 18: `workqueueSize` is not used anywhere in manager or reconciler

controllers/instnace_binding_controller_manager.go
```diff
@@ -0,0 +1,87 @@
...
+	r.sisbControllerMu.Lock()
+	defer r.sisbControllerMu.Unlock()
+	logger := log.Log
+
+	if r.enabled {
+		return
+	}
+	mgr, err := ctrl.NewManager(r.cfg, ctrl.Options{
+		Scheme:             r.Scheme,
+		MetricsBindAddress: "0",
```
szwedm, line 48: maybe we should disable `HealthProbeBindAddress` as well?

szwedm (approved):
/lgtm

PR #274, opened 2023-04-27
Add raw Kubernetes YAML handling scenario for e2e-test

### LINKED ISSUE #273: Failing run-e2e-test
The value found in manifest.json `"mediaType: "application/gzip"` for CustomResourceDefinition changed to `"application/octet-stream`. This caused the `run-e2e-test` pipeline to [fail](https://github.com/kyma-project/btp-manager/actions/runs/4807608875/jobs/8556689080) because we are looking for the type [`application/gzip`](https://github.com/kyma-project/btp-manager/blob/6e4a55e174ef09ad5c91d4f2f29cc5f7c13e9063/scripts/run_module_image.sh#L37). The pipeline should be able to handle the old and new mediaType

The idea is to add a mediaType check and add an alternative path if no `mediaType: "application/gzip"` is detected. We could use `jq` to check this, and then also to detect the rendered.yaml file containing `kind: CustomResourceDefinition` for the alternative path.

MarekMichali:
**Description**

The e2e-test had no way to handle raw Kubernetes yaml, only helm charts. This PR adds this feature

- Add raw Kubernetes YAML handling scenario for e2e-test

**Related issue(s)**

Resolves #273

### DIFF
--- a/scripts/run_module_image.sh
+++ b/scripts/run_module_image.sh
@@ -36,12 +36,30 @@ skopeo copy ${TLS_OPTIONS} docker://${IMAGE_NAME} dir:${TARGET_DIRECTORY}
 
 FILENAME=$(cat ${TARGET_DIRECTORY}/manifest.json  | jq -c '.layers[] | select(.mediaType=="application/gzip").digest[7:]' | tr -d \")
 
-echo -e "\n--- Extracting resources from file:" ${FILENAME}
+# if no application/gzip files found that means that we have a raw Kubernetes YAML file and we need to identify it
+if [ -z "${FILENAME}" ]
+then
+  POTENTIAL_YAML_FILES=$(cat ${TARGET_DIRECTORY}/manifest.json  | jq -c '.layers[] | select(.mediaType=="application/octet-stream").digest[7:]' | tr -d \")
+  POTENTIAL_YAMLS_ARRAY=($POTENTIAL_YAML_FILES)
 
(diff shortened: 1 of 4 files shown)


scripts/run_module_image.sh
```diff
@@ -36,12 +36,30 @@ skopeo copy ${TLS_OPTIONS} docker://${IMAGE_NAME} dir:${TARGET_DIRECTORY}
...
+then
+  POTENTIAL_YAML_FILES=$(cat ${TARGET_DIRECTORY}/manifest.json  | jq -c '.layers[] | select(.mediaType=="application/octet-stream").digest[7:]' | tr -d \")
+  POTENTIAL_YAMLS_ARRAY=($POTENTIAL_YAML_FILES)
+
+  for RAW_K8S_YAML in "${POTENTIAL_YAMLS_ARRAY[@]}"
+  do
+    KIND_VALUE=$(cat ${TARGET_DIRECTORY}/${RAW_K8S_YAML} | yq e '.kind' -)    
+    if echo "$KIND_VALUE" | grep -q "CustomResourceDefinition"; 
+    then
+      echo -e "\n--- Installing BTP Manager in ${NAMESPACE} namespace"
```
jaroslaw-pieszka, line 50: Please augment message with information about whether we install using helm or kubectl apply

jaroslaw-pieszka (approved):

PR #297, opened 2023-05-18
metrics v1

ukff:
- move conditions to internal package
- add simple metrics to track number of occured states of operator


controllers/utils_test.go
```diff
@@ -57,6 +59,7 @@ const (
 	priorityClassYamlPath = "testdata/test-priorityclass.yaml"
 	k8sOpsTimeout         = time.Second * 3
 	k8sOpsPollingInterval = time.Millisecond * 200
+	readyType             = "Ready"
```
ralikio, line 62: Can we used ReadyType from `conditions` package?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+	reasonsCounters := make(map[conditions.Reason]prometheus.Counter, len(conditions.Reasons))
+	for reason, metadata := range conditions.Reasons {
+		counter := prometheus.NewCounter(prometheus.CounterOpts{
+			Name:        string(reason),
+			ConstLabels: prometheus.Labels{"state": string(metadata.State)},
+		})
+		reasonsCounters[reason] = counter
+		metrics.Registry.MustRegister(counter)
+	}
+	m.ReasonCounters = reasonsCounters
```
ralikio, line 26: Can we used `m.ReasonCounters` directly in the assignment and the for loop?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func InitMetrics() *Metrics {
```
ralikio, line 29: What do you think about using `New*` methods?

internal/metrics/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+type Metrics struct {
+	mutex          sync.Mutex
+	ReasonCounters map[conditions.Reason]prometheus.Counter
+}
+
+func (m *Metrics) registerMetrics() {
```
ralikio, line 16: What do you think about using mutex in this method as well?

controllers/serviceinstance_controller.go
```diff
@@ -60,19 +65,20 @@ func (r *ServiceInstanceReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
-func (r *ServiceInstanceReconciler) UpdateBtpOperatorStatus(ctx context.Context, cr *v1alpha1.BtpOperator, newState types.State, reason Reason, message string) error {
+func (r *ServiceInstanceReconciler) UpdateBtpOperatorStatus(ctx context.Context, cr *v1alpha1.BtpOperator, newState types.State, reason conditions.Reason, message string) error {
 	cr.Status.WithState(newState)
-	newCondition := ConditionFromExistingReason(reason, message)
+	newCondition := conditions.ConditionFromExistingReason(reason, message)
 	if newCondition != nil {
-		SetStatusCondition(&cr.Status.Conditions, *newCondition)
+		conditions.SetStatusCondition(&cr.Status.Conditions, *newCondition)
 	}
+	r.metrics.IncreaseReasonCounter(reason)
```
ralikio, line 81: I am not sure if using the same `metrics` might block both reconciliations. Is it possible to populate our controllers with separate metrics instances and avoid a mutex at all?

ralikio (changes requested):

ralikio (approved):

PR #318, opened 2023-06-06
Reduce the risk of  OOM

jaroslaw-pieszka:
**Description**

- RateLimiter used
- Disabling SI/SB controller earlier
- script for memory stress test

**Related issue(s)**

### DIFF
--- a/scripts/testing/stress-mem.sh
+++ b/scripts/testing/stress-mem.sh
@@ -14,7 +14,7 @@ LIFE_SPAN=${2-60}
 
 echo -e "\n---Installing BTP operator"
 kubectl apply -f ${YAML_DIR}/e2e-test-btpoperator.yaml
-kubectl label btpoperators e2e-test-btpoperator force-delete=true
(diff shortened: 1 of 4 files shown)


scripts/testing/stress-mem.sh
```diff
@@ -0,0 +1,53 @@
...
+#                       - number of seconds to let them be before btp-operator is deleted - default 60
+# Example:
+#           ./stress-mem.sh 100 30
+
+N=${1-500}
+YAML_DIR=./scripts/testing/yaml
+LIFE_SPAN=${2-60}
+
+echo -e "\n---Installing BTP operator"
+kubectl apply -f - <<EOF
```
szwedm, line 16: we can apply `e2e-test-btpoperator.yaml` and then label it with `kubectl label`, like this: `kubectl label -f e2e-test-btpoperator.yaml force-delete=true`

jaroslaw-pieszka: In the next commit.

szwedm (approved):
/lgtm

PR #336, opened 2023-07-05
Update resource during reconcile if resource exists

jaroslaw-pieszka:
**Description**

We used to use patch with apply regardless of whether resource existed or not. This approach caused to leave untouched labels and annotations attached by `reconciler`.

- If resource already exists we use update, if not we use patch with apply (as previously)

**Related issue(s)**


szwedm (approved):
/lgtm

PR #342, opened 2023-07-11
Fix handling multiple BtpOperator CRs

### LINKED ISSUE #335: Multiple BtpOperator CRs restore scenario doesn't work
**Description**
We see some BTP Manager instances have one BtpOperator CR instance in the Error/OlderCRExists. Then after around 10h this was finally reconciled and it went into the Ready state as expected.

**AC**
- [x] 1) Write a unit test that proves that we have one BtpOperator CR with state OlderCRExists
- [x] 2) Find and place which causes an issue and fix it
- [x] ~~3) Consider using global SyncPeriod, to avoid such issues in the future~~ Use reconcile.Result instead.

ukff:
Handle multiple BtpOperator CRs when current leader BtpOperator CR (currently responsible for the module reconciliation) is deleted. Add unit tests.

Resolves #335

### DIFF
--- a/controllers/btpoperator_controller_cr_rotation_test.go
+++ b/controllers/btpoperator_controller_cr_rotation_test.go
@@ -36,12 +36,11 @@ var _ = Describe("BTP Operator CR leader replacement", func() {
 			for _, cr := range btpOperators.Items {
 				btpOp := cr.DeepCopy()
 				Expect(k8sClient.Delete(ctx, btpOp)).To(Succeed())
+				Eventually(updateCh).Should(Receive(matchDeleted()))
 			}
-			Eventually(func() (bool, error) {
-				err := k8sClient.List(ctx, btpOperators)
-				return len(btpOperators.Items) == 0, err
-			}).WithTimeout(k8sOpsTimeout).WithPolling(k8sOpsPollingInterval).Should(BeTrue())
 		}
+		Expect(k8sClient.List(ctx, btpOperators)).To(Succeed())
+		Expect(len(btpOperators.Items)).To(BeEquivalentTo(0))
 		deleteSecret := &corev1.Secret{}
 		Expect(k8sClient.Get(ctx, client.ObjectKey{Namespace: kymaNamespace, Name: SecretName}, deleteSecret)).To(Succeed())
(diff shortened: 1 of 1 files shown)


controllers/btpoperator_controller.go
```diff
@@ -199,41 +207,65 @@ func (r *BtpOperatorReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	logger.Info(fmt.Sprintf("Found %d existing BtpOperators", len(existingBtpOperators.Items)))
+	oldestCr := r.getOldestCR(existingBtpOperators)
+	if oldestCr.GetStatus().State == v1alpha1.StateError && oldestCr.IsReasonStringEqual(string(conditions.OlderCRExists)) {
+		if err := r.UpdateBtpOperatorStatus(ctx, oldestCr, v1alpha1.StateProcessing, conditions.Processing,
+			fmt.Sprintf("%s is the new leader", oldestCr.GetName())); err != nil {
+			logger.Error(err, fmt.Sprintf("unable to set %s BtpOperator CR as the new leader", oldestCr.GetName()))
+			return ctrl.Result{}, err
+		}
+	}
+	logger.Info(fmt.Sprintf("%s BtpOperator is the new leader", oldestCr.GetName()))
```
jaroslaw-pieszka, line 256: We reach this point not only when we successfully updated the oldest CR and set is as a new leader but even if `oldestCr.GetStatus().State == v1alpha1.StateError` but the reason is other than `conditions.OlderCRExists`.

szwedm: That's correct, the oldest BtpOperator will be picked up as a new leader and it doesn't have to be in Error State with OlderCRExists condition. If condition at [L249](https://github.com/kyma-project/btp-manager/pull/342/files#diff-c40c410af64e0ab33ab23c7fb7cd4bfdf0c7a4fe7c90b5d798a7938303c73edbR249) is required to avoid unnecessary, redundant reconciliations. BtpOperators with Statuses other than Error State, OlderCRExists Condition should be reconciled automatically.

jaroslaw-pieszka (approved):

Issue #353, opened 2023-07-20
Adjust the diagrams in the `btp-manager` documentation to the content guidelines


IwonaLanger:
- Make necessary changes in the following diagrams:
  - [x] 1) provisioning
  - [x] 2) deprovisioning
  - [x] 3) certs

- Remove the outdated diagram
  - [x] 4) update

- Make a diagram using the recommended tool (diagrams.net)
  - [x] 5) creating a release

PR #354, opened 2023-07-21
Adjust operations and certs diagrams

IwonaLanger:
**Description**

- Remove the outdated "update" diagram
- Update "provisioning", "deprovisioning", and "certs" diagrams
- Add the "release" diagram created in diagram.net

**Related issue(s)**
#353


docs/assets/deprovisioning.svg
nataliasitko, line 1: [Our guidelines suggest](https://github.com/kyma-project/community/blob/main/docs/guidelines/content-guidelines/09-diagrams.md#background) using mint green (HEX: #DEF2DD) for subsidiary environments. Would it make sense to replace those white "Hard delete" and "Soft delete" rectangles with rounded rectangles in that mint green color?

nataliasitko: Also, I don't know if we should use sentence case for the texts in diagrams or begin with a lowercase letter. It seems that we're not consistent about that in Kyma

IwonaLanger: > [Our guidelines suggest](https://github.com/kyma-project/community/blob/main/docs/guidelines/content-guidelines/09-diagrams.md#background) using mint green (HEX: #DEF2DD) for subsidiary environments. Would it make sense to replace those white "Hard delete" and "Soft delete" rectangles with rounded rectangles in that mint green color?

Now that you mention it, it makes perfect sense :) I'll change the color.

IwonaLanger: > Also, I don't know if we should use sentence case for texts in diagrams or begin with a lowercase letter. It seems that we're not consistent about that within Kyma

Same thoughts exactly. I'll have another look and adapt the prevailing approach.

nataliasitko (approved):

docs/assets/deprovisioning.svg
IwonaLanger: > Thanks, I can create an issue to update the diagrams that use the less common approach :)

Good idea :)

piotrmiskiewicz (approved):

PR #370, opened 2023-08-07
Add certs regenerations counter metric

MarekMichali:
**Description**

- ...
- ...
- ...

**Related issue(s)**

### DIFF
--- a/controllers/btpoperator_controller.go
+++ b/controllers/btpoperator_controller.go
@@ -1480,6 +1480,7 @@ func (r *BtpOperatorReconciler) prepareCertificatesReconciliationData(ctx contex
 		return err
 	}
 	if certificatesRegenerationDone {
+		r.metrics.IncreaseCertsRegenerationsCounter()
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/btpoperator_controller.go
```diff
@@ -1480,6 +1480,7 @@ func (r *BtpOperatorReconciler) prepareCertificatesReconciliationData(ctx contex
 		return err
 	}
 	if certificatesRegenerationDone {
+		r.metrics.IncreaseCertsRegenerationsCounter()
```
jaroslaw-pieszka, line 1483: Why don't we instrument the `doFullCertificatesRegeneration` with metrics.

MarekMichali: There is also a `doPartialCertificatesRegeneration` that is independent of `doFullCertificatesRegeneration`

MarekMichali: I could move this to `ensureCertificatesExists` before the return statements, where we signal that a regen has occurred

jaroslaw-pieszka: Leave it as it is please.

jaroslaw-pieszka (approved):

PR #426, opened 2023-09-06
Add sec-scanner-bump to release workflow

MarekMichali:
**Description**

The new `Bump sec-scanners-config` step in the release workflow creates a PR with the new image version in sec-scanner-config.yaml and waits for the PR to be merged or closed. Approval of the PR is done manually by the codeowners.

The `Bump sec-scanners-config` step can be skipped by unchecking a checkbox when setting up the release.

- Add a step to create a PR with a new sec-scanner-config.yaml 
- Add a check to `Validate release` if the tag already exists

**Related issue(s)**


scripts/create_scan_config.sh
```diff
@@ -17,8 +17,9 @@ IMAGE=${1}
 echo "Creating security scan configuration file:"
 cat <<EOF | tee ${FILENAME}
 module-name: btp-operator
+rc-tag: ${IMAGE}
```
jaroslaw-pieszka, line 20: Shouldn't we rename it to different name if the meaning changes.

scripts/create_scan_config.sh
```diff
@@ -17,8 +17,9 @@ IMAGE=${1}
...
 protecode:
-  - ${IMAGE}
+  - europe-docker.pkg.dev/kyma-project/prod/btp-manager:${IMAGE}
```
jaroslaw-pieszka, line 22: Is there a simple way to pass it to the script (to reduce the number of places where repository location is hard-wired). If this requires too much hassle, leave it as it is.

scripts/create_scan_config.sh
MarekMichali: I will rename it to `TAG`

scripts/create_scan_config.sh
MarekMichali: I could pass the path to the image as an argument. Also the `{IMAGE}` will be replaced with `{TAG}`

MarekMichali: We have already an env `IMAGE_REPO: europe-docker.pkg.dev/kyma-project/prod/btp-manager` so it doesn't requieres too much hassle.

jaroslaw-pieszka (approved):

PR #472, opened 2023-10-20
New promotion with fix and updated comments

jaroslaw-pieszka:
**Description**

- Explicit head in `gh pr create` CLI call to prevent erroneous behaviour when the command inaccurately calculates head 
- Comments specifying required variables

**Related issue(s)**


jaroslaw-pieszka:
Commits without user do not allow to merge.

PR #483, opened 2023-10-26
Fix commit refs during release

MarekMichali:
**Description**

If we did a commit to main while the workflow was running, the latest commit was not recognized because the ref in github env is set on workflow dispatch.

- Save latest commit ref to custom env

**Related issue(s)**


szwedm (approved):
/lgtm

PR #538, opened 2023-12-21
Update deps 211223 

ukff:
**Description**

- ...
- ...
- ...

**Related issue(s)**


Dockerfile
```diff
@@ -12,13 +12,16 @@ RUN go mod download
 # Copy the go source
 COPY . ./
 
+RUN go mod tidy
```
ralikio, line 15: Because mod tidy updates dependencies we might get to the point where automatic image builds results in dependencies from container being different than dependencies in the repository.

ralikio (approved):

PR #555, opened 2024-01-03
Add alert to slack if auto update job will fail

ukff:


.github/workflows/auto-update.yml
```diff
@@ -59,3 +60,12 @@ jobs:
           else
             scripts/update/create_pr.sh
           fi
+
+      - name: Post to a Slack channel failure
```
jaroslaw-pieszka, line 64:
```suggestion
      - name: Notify slack channel about failure
```

MarekMichali (approved):

PR #649, opened 2024-04-05
Service Manager UI

ukff:
- add UI for Service Offerings (Marketplace) and for Service Instances

### DIFF
--- a/internal/service-manager/client.go
+++ b/internal/service-manager/client.go
@@ -41,10 +41,7 @@ type Client struct {
 	smURL          string
 }
 
-func NewClient(
-	ctx context.Context, logger *slog.Logger,
-	secretProvider clusterobject.NamespacedProvider[*corev1.Secret],
-) *Client {
+func NewClient(ctx context.Context, logger *slog.Logger, secretProvider clusterobject.NamespacedProvider[*corev1.Secret]) *Client {
 	return &Client{
 		ctx:            ctx,
 		logger:         logger.With("component", componentName),
--- a/internal/service-manager/client_test.go
+++ b/internal/service-manager/client_test.go
@@ -149,10 +149,6 @@ func (p *fakeSecretProvider) GetByNameAndNamespace(ctx context.Context, name, na
 	return nil, fmt.Errorf("secret not found")
 }
 
-func (p *fakeSecretProvider) All(ctx context.Context) (*corev1.SecretList, error) {
-	return nil, nil
-}
-
 func (p *fakeSecretProvider) clean() {
 	p.secrets = make([]*corev1.Secret, 0)
 }


Makefile
```diff
@@ -199,3 +199,28 @@ go-lint: go-lint-install ## linter config in file at root of project -> '.golang
...
+app: ## Run the app
+	go run main.go
+
+.PHONY: ui
+ui: ## Run the ui
+	cd ui && npm install && npm start
+
+.PHONE: webapp
+webapp: ## Run App and UI
+	 @$(MAKE) -j ui app
```
szwedm, line 213: Can you explain the case to use this rule?
https://www.gnu.org/software/make/manual/make.html#Overriding-Makefiles

Makefile
```diff
@@ -199,3 +199,28 @@ go-lint: go-lint-install ## linter config in file at root of project -> '.golang
...
+
+.PHONY: clean-ports
```
szwedm, line 215: Not required if random available ports are assigned in the app and stored in environment variables.

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.22.3
 
 require (
 	github.com/go-logr/logr v1.4.1
+	github.com/gorilla/mux v1.8.1
```
szwedm, line 7: You can use routing from std lib http package instead:
https://go.dev/doc/go1.22#enhanced_routing_patterns

internal/api/api.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/gorilla/mux"
+	"github.com/kyma-project/btp-manager/internal/api/vm"
+	clusterobject "github.com/kyma-project/btp-manager/internal/cluster-object"
+	servicemanager "github.com/kyma-project/btp-manager/internal/service-manager"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+type API struct {
+	cluster                 *client.Client
```
szwedm, line 20: Not required in the struct, you can pass already created SM client from main func.

internal/api/api.go
```diff
@@ -0,0 +1,124 @@
...
+	namespaceProvider       *clusterobject.NamespaceProvider
```
szwedm, line 21: Not required in the struct, you can pass already created SM client from main func.

internal/api/api.go
```diff
@@ -0,0 +1,124 @@
...
+	serviceInstanceProvider *clusterobject.ServiceInstanceProvider
```
szwedm, line 22: Reading Service Instances from the cluster is out of scope, `ServiceInstanceProvider` is a dependency to SM client  for retrieving secrets names referenced in SI resources. SM client will provide Service Instances information from Service Manager.

internal/api/api.go
```diff
@@ -0,0 +1,124 @@
...
+	secretProvider := clusterobject.NewSecretProvider(k8s, namespaceProvider, serviceInstanceProvider, log)
+	serviceManager := servicemanager.NewClient(context.Background(), log, secretProvider)
+
+	return &API{
+		cluster:                 &k8s,
+		namespaceProvider:       namespaceProvider,
+		serviceInstanceProvider: serviceInstanceProvider,
+		secretProvider:          secretProvider,
+		serviceManager:          serviceManager,
+	}, nil
```
szwedm, line 48: I think that all the api needs is SecretProvider and SM client which can be created in the main func and then passed to the constructor.

internal/api/api.go
```diff
@@ -0,0 +1,124 @@
...
+	if shouldReturnError(writer, err) {
+		return
+	}
+	_, err = (*writer).Write(response)
+	if shouldReturnError(writer, err) {
+		return
+	}
+}
+
+func shouldReturnError(writer *http.ResponseWriter, err error) bool {
```
szwedm, line 114: The func name is confusing, looks like an assert func for tests.

szwedm: Why use pointer to the interface?
https://pkg.go.dev/net/http@go1.22.3#ResponseWriter

internal/api/vm/converters.go
```diff
@@ -0,0 +1,71 @@
+package vm
```
szwedm, line 1: What is `vm` package?

internal/api/vm/converters.go
```diff
@@ -0,0 +1,71 @@
...
+				ImageUrl:    imageUrl,
+				DisplayName: displayName,
+			},
+		}
+		serviceOfferings.Items = append(serviceOfferings.Items, offering)
+	}
+	return serviceOfferings
+}
+
+func ToServiceInstancesVM(serviceInstances []unstructured.Unstructured) ServiceInstances {
```
szwedm, line 48: Service Instances will be provided by SM client in an appropriate struct then func will need adjustments. I believe it's better to wait for SM client functionality than work on unstructured objects.

main.go
```diff
@@ -41,6 +45,10 @@ import (
 	//+kubebuilder:scaffold:imports
 )
 
+const (
+	apiPort = 3006
```
szwedm, line 49: I think it would be better to assign the port in a values.yaml file and then read environment variable holding the port number.

main.go
```diff
@@ -180,8 +188,27 @@ func setupSMClient(restCfg *rest.Config, signalCtx context.Context) *serviceMana
...
+	api, err := api.NewAPI()
+	if err != nil {
+		return fmt.Errorf("unable to create API: %w", err)
+	}
+	mux := mux.NewRouter()
+	mux.HandleFunc("/api/list-secrets", api.ListSecrets).Methods("GET")
+	mux.HandleFunc("/api/list-service-instances", api.ListServiceInstances).Methods("GET")
+	mux.HandleFunc("/api/get-service-instance/{id}", api.GetServiceInstance).Methods("GET")
+	mux.HandleFunc("/api/list-service-offerings/{namespace}/{name}", api.ListServiceOfferings).Methods("GET")
+	mux.HandleFunc("/api/get-service-offering/{id}", api.GetServiceOffering).Methods("GET")
```
szwedm, line 212: Take a look at https://pkg.go.dev/net/http@go1.22.3#ServeMux

main.go
```diff
@@ -180,8 +188,27 @@ func setupSMClient(restCfg *rest.Config, signalCtx context.Context) *serviceMana
...
+	return http.ListenAndServe(fmt.Sprintf(":%d", apiPort), mux)
```
szwedm, line 213: `ListenAndServe` will block main goroutine.

szwedm (commented):
Please take a look at my comments.

PR #789, opened 2024-08-07
Increase RSA key size

KsaweryZietara:
**Description**

- increase RSA key size to 4096,
- bump deployment limit memory to 500MB,
- bump deployment limit CPU to 1.

**Related issue(s)**
See also #791


PK85:
Question, when changes will be reflected in the module-resources?

PK85 (approved):

Issue #442, opened 2023-09-20
[POC] Catalog Views and integration with Service Manager


PK85:
**Description**

Provide new UI for BTP Operator in the Busola where users can see Service Catalog. For now it could be separate application, if installed in the Kyma cluster should already work. In the future probably it will be a part of the btp-manager module.

**Reason**

We want to make `BTP Operator` user's lives easier and more convenient where they can see all of the available services in the Busola and then just provision some of them.

**AC**

- [x] 1) Direct integration with the Service Manager
  - [x] 1.1) User can use default credentials  + can select secret from any namespace (SM credentials are from different SA and then could see different instances and service offerings)
https://github.com/kyma-project/btp-manager/pull/655 https://github.com/kyma-project/btp-manager/pull/674 https://github.com/kyma-project/btp-manager/pull/703
  - [x] 1.2) user can see a list of all service offerings from Service Manager from given SA (SM credentials are bound to given SA)
 https://github.com/kyma-project/btp-manager/pull/717
  - [x] 1.3) User can see a list of all existing instances from Service Manager from given SA (SM credentials are bound to given SA) https://github.com/kyma-project/btp-manager/pull/742, https://github.com/kyma-project/btp-manager/pull/735, https://github.com/kyma-project/btp-manager/pull/746 https://github.com/kyma-project/btp-manager/pull/758
  - [x] 1.4) User can create a service binding on Service Manager side which results as a secret in the selected namespace in the Kyma environment https://github.com/kyma-project/btp-manager/pull/735 https://github.com/kyma-project/btp-manager/pull/757
  - [x] 1.5) User can delete a service binding on Service Manager side which invalidates given secret (probably we need to delete corresponding secret) https://github.com/kyma-project/btp-manager/pull/757 https://github.com/kyma-project/btp-manager/pull/772
- [ ] 2) New BTP Operator UI in the Busola
  - [ ] 2.1) check with Busola team, on how to embed our own web page (We should have a view in the namespace + in the overview taking into account all namespaces data)

- [ ] Think in future of making it mobile - friendly

**Links**

- check smctl what calls are executed
- https://github.com/Peripli/service-manager
- https://help.sap.com/docs/service-manager/sap-service-manager/sap-service-manager

PK85:
Reference sharing:
```
apiVersion: services.cloud.sap.com/v1
kind: ServiceInstance
metadata:
  name: kyma-demo-instance-123
  namespace: develope
spec:
[...]

PR #809, opened 2024-08-14
Load secrets onClick

MarekMichali:
**Description**

Load secrets onClick "Select secret" to not force the user to reload the page when a new secret is added.

- Load secrets onClick
- Refresh views

**Related issue(s)**

See #442


szwedm (approved):
/lgtm

PR #830, opened 2024-08-21
404 on Refresh Fix

ralikio:
**Description**

The following solution is adaption of https://stackoverflow.com/a/64687181 to provide default route to ui app if the file server does not contain given path. The file server returns 404 path if requested path does not correspond to any files in its filesystem. Because of that while routing request to our single page app (e.g. path "/instances/{id}") the request was handled by file server and not by the app itself (which corresponds to `index.html`). The solution routes existing paths, like "/static/*", to the file server and non existing, like "/instances/{id}", to the app frontend.

**Related issue(s)**

#442

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -8,6 +8,8 @@ import (
 	"log"
 	"log/slog"
 	"net/http"
+	"os"
+	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


szwedm (approved):
/lgtm

PR #940, opened 2024-12-24
Fix E2E test on release

MarekMichali:
**Description**

E2E tests on real credentials don't use the btp-manager configmap. This PR creates the configmap when needed

- Create btp-manager configmap only when needed

**Related issue(s)**


szwedm (approved):
/lgtm

PR #974, opened 2025-02-06
Auto merging bump PR for release workflow

jaroslaw-pieszka:
**Description**

Features we use in KEB release process

- helper scripts
- checks during release to detect changes introduced during the release

**Related issue(s)**


.github/workflows/create-release.yaml
```diff
@@ -224,6 +165,107 @@ jobs:
       image-registry: europe-docker.pkg.dev/kyma-project/prod/btp-manager
       image-tag: ${{ github.event.inputs.name }}
 
+  bump-sec-scanners-config:
+    name: Bump sec-scanners-config
+    needs: validate-release
+    runs-on: ubuntu-latest
```
MarekMichali, line 171: The `needs` sections require adjustments here and in the `build-image` section.

.github/workflows/create-release.yaml
```diff
@@ -262,12 +316,24 @@ jobs:
...
+            exit 1 
+          fi
+
     outputs:
       release_id: ${{ steps.create-draft.outputs.release_id }}
 
   publish-release:
     name: Publish release
-    needs: [create-draft, run-unit-tests, run-e2e-tests, run-e2e-upgrade-tests, run-stress-tests, run-e2e-upgrade-while-deleting-tests]
+    needs: [create-draft, bump-sec-scanners-config]
```
MarekMichali, line 336: This `bump-sec-scanners-config` unlike in KEB is not needed here because we don't checkout here to the commit created in the bump step.

MarekMichali: In KEB we create a helm package and this is why we need to checkout to the latest version. Here we run only the publish where only the draft ID is needed.

MarekMichali (approved):

PR #1249, opened 2025-11-24
Update the BTP Manager Operations doc for NS2

IwonaLanger:
**Description**

- Revised the document.
- Updated the diagrams.
- Replaced relative links with absolute ones, so they render correctly when the doc is uploaded to the Help Portal.
- Removed the `_sidebar.md` file; it's been replaced with `_sidebar.ts`


mmitoraj (commented):
Shouldn't the CR states and/or types in the table be in the font code? How about the condition reason?

mmitoraj (approved):

mmitoraj (approved):

PR #1311, opened 2026-01-14
Update user documentation

IwonaLanger:
**Description**

- Updates
- Language and formatting fixes
- Adjust Network Policies to be published on Help Portal

### DIFF
--- a/docs/user/03-15-network-policies.md
+++ b/docs/user/03-15-network-policies.md
@@ -4,7 +4,7 @@ Learn about the network policies for the SAP BTP Operator module and how to mana
 
 ## Default Network Policies
 
-To increase security, the SAP BTP Operator module creates the following network policies that control traffic to and from the SAP BTP Operator module Pods.
+To increase security, the SAP BTP Operator module creates the following network policies that control traffic to and from its Pods:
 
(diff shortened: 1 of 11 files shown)


docs/user/03-15-network-policies.md
```diff
@@ -1,37 +1,35 @@
 # Network Policies
 
-The SAP BTP Operator module can create network policies to control traffic for the SAP BTP service operator Pods. By default, network policies are enabled.
+Learn about the network policies for the SAP BTP Operator module and how to manage them.
+
+## Default Network Policies
+
+By default, the SAP BTP Operator module creates the following network policies to control traffic to and from the SAP BTP service operator Pods. <!--should we mention anything about the security-related purpose of the network policies?-->
```
MarekMichali: We control traffic for both the btp-manager Pod and SAP BTP service operator Pod. Maybe we could change it to just something like this:

> to control traffic to and from the SAP BTP Operator module Pods.

We can mention that this is used to limit the ingress/egress traffic to only what's necessary, which increases security by restricting potential attacks. But on the other hand, NetworkPolicies are basic Kubernetes resources and I don't think that we explain things like ConfigMaps. The user should know what a NetworkPolicy is from Kubernetes docs.

Issue #1330, opened 2026-01-27
Extensible resources application to the cluster


szwedm:
The manager should provide the possibility to add more applicable resources as an extension in the process for applying the [module resources](https://github.com/kyma-project/btp-manager/tree/main/module-resources). Users should have an option to include/exclude the extra resources. The resources added as the extension should underway the same common preparation steps as the module resources (setting namespace, labels, etc.). The [manager resources](https://github.com/kyma-project/btp-manager/tree/0ff1ce33c171fe2d63aab2527b5b5666714f56c8/manager-resources) directory stores manifests of the extra resources for the application process.

Network policies are a good example of the extra resources for application to the cluster:
[manifests](https://github.com/kyma-project/btp-manager/blob/0ff1ce33c171fe2d63aab2527b5b5666714f56c8/manager-resources/network-policies/network-policies.yaml)
[adding/deleting mechanism in the reconciliation process](https://github.com/kyma-project/btp-manager/blob/0ff1ce33c171fe2d63aab2527b5b5666714f56c8/controllers/btpoperator_controller.go#L525-L536)

**AC:**
- [ ] 1. The manager applies the extra resources when the user selects the option to include extra resources
- [ ] 2. The manager deletes the extra resources when the user selects the option to exclude extra resources
- [ ] 3. Unit tests for both scenarios

KsaweryZietara:
Blocked until the changes from `refactor/separate-concerns` are merged into `main`.

MarekMichali:
We decided to not implement it. If we have more resources like the network policies, we will consider a generic approach.

PR #1350, opened 2026-02-10
Extensible resources application to the cluster

KsaweryZietara:
**Description**

- Add a Resorce Manager.
- Implement NetworkPolicies as a managed resource, controlled via the `DisableNetworkPolicies` annotation.
- Support future extension with additional resource types through the Resource interface.

**Related issue(s)**
See also #1330

### DIFF
--- a/internal/manager/managerresource/manager.go
+++ b/internal/manager/managerresource/manager.go
@@ -65,7 +65,6 @@ func (m *Manager) ResourcesToDelete(ctx context.Context) ([]client.Object, error
 		if resource.Enabled() {
 			continue
 		}
-
 		logger.Info(fmt.Sprintf("%s disabled, preparing existing resources for removal", resource.Name()))
 		resources = append(resources, resource.Object())
 	}
--- a/internal/manager/managerresource/manager.go
+++ b/internal/manager/managerresource/manager.go
@@ -44,7 +44,7 @@ func NewManager(client client.Client, scheme *runtime.Scheme, resources []Resour
 	}
 }
 
-func (m *Manager) createManagerResources(ctx context.Context) error {
+func (m *Manager) CreateManagerResources(ctx context.Context) error {
 	logger := log.FromContext(ctx)
 
 	btpOperator := &v1alpha1.BtpOperator{}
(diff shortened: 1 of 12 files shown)


internal/manager/managerresource/manager.go
```diff
@@ -0,0 +1,136 @@
...
+func NewManager(client client.Client, scheme *runtime.Scheme, resources []Resource) *Manager {
+	return &Manager{
+		client:          client,
+		scheme:          scheme,
+		manifestHandler: &manifest.Handler{Scheme: scheme},
+		resources:       resources,
+	}
+}
+
+func (m *Manager) createManagerResources(ctx context.Context) error {
```
szwedm: Private method. We won't have access to it from the module resource manager.

internal/manager/managerresource/manager.go
```diff
@@ -0,0 +1,136 @@
...
+	logger := log.FromContext(ctx)
+
+	btpOperator := &v1alpha1.BtpOperator{}
+	if err := m.client.Get(ctx, client.ObjectKey{Namespace: config.KymaSystemNamespaceName, Name: config.BtpOperatorCrName}, btpOperator); err != nil {
+		return fmt.Errorf("failed to get BTP Operator CR: %w", err)
+	}
```
szwedm: It's BtpOperatorReconciler's responsibility to get the BtpOperator CR. I think the Resource Manager should work on the CR which already triggered the reconciliation in BtpOperatorReconciler.

internal/manager/managerresource/manager.go
```diff
@@ -0,0 +1,136 @@
...
+			unstructuredObjects, err := m.manifestHandler.ObjectsToUnstructured(objects)
+			if err != nil {
+				return fmt.Errorf("while converting to unstructured: %w", err)
+			}
+
+			logger.Info(fmt.Sprintf("creating %d %s", len(objects), resource.Name()))
+			err = m.createOrUpdateResources(ctx, unstructuredObjects)
+			if err != nil {
+				return fmt.Errorf("failed to create %s: %w", resource.Name(), err)
+			}
```
szwedm: The component should prepare extra resources to apply (if enabled) as extensions to the module resources from the Module Resource Manager. The extra resources should undergo the same preparation process as the module resources (setting labels, namespaces, etc.). I think creating, updating, deleting responsibilities should stay in the Module Resource Manager only. This way we will reduce operations in the cluster through K8s client to only one component.

KsaweryZietara: The `ResourcesToCreate` function returns a list of `unstructured.Unstructured`, and `ResourcesToDelete` returns a list of `client.Object`. Neither function currently performs any Kubernetes operations.

PR #1440, opened 2026-04-22
Fix flaky test

piotrek1459:
**Description**

Fix three flaky tests caused by race conditions in the shared envtest cluster.

- internal/ymlutils/renamer.go — AddSuffixToNameInManifests now accepts an optional excludeNames ...string parameter. Names in the exclusion list are never renamed, keeping stable operand resource names (sap-btp-operator-config, sap-btp-service-operator, sap-btp-operator-controller-manager) intact — matching real upgrade behavior where these names never change.

controllers/btpoperator_controller_updating_test.go — Three fixes:

- All three stable names (ConfigMap, Secret, Deployment) are passed as excludes to AddSuffixToNameInManifests, so config.DeploymentName is no longer mutated during the test, eliminating a race window between the YAML rename and the config variable update.
- Eventually(workqueueSize == 0) moved to the top of both rename test cases — ensures the background reconciler completes with a consistent original state before any YAML modifications, guaranteeing a 15-minute quiet window (the RequeueAfter from a successful reconcile) rather than a 5ms one (exponential backoff after a failed reconcile).
assertResourcesRemoval filters out stable-name resources that are intentionally not renamed and therefore still exist after reconciliation.
- controllers/btpoperator_controller_network_policies_test.go — cleanupNetworkPolicies added to BeforeEach. The previous AfterEach cleanup fired on CR deletion, not on child resource removal, so NetworkPolicies could linger into the next test.

- controllers/btpoperator_controller_configuration_test.go — Replaced Eventually(Create(configMap)) with Eventually(StateReady) + Get + Update. By the time the CR reaches Ready, the reconciler has already created the ConfigMap; trying to create it again deadlocked.

No production code was modified.

**Related issue(s)**


jaroslaw-pieszka (changes requested):
Do not change production code.

jaroslaw-pieszka (dismissed):

controllers/config/handler.go
```diff
@@ -27,10 +27,12 @@ const (
...
 var (
-	ChartNamespace = "kyma-system"
-	SecretName     = "sap-btp-manager"
-	ConfigName     = "sap-btp-manager"
-	DeploymentName = "sap-btp-operator-controller-manager"
+	ChartNamespace       = "kyma-system"
+	SecretName           = "sap-btp-manager"
+	ConfigName           = "sap-btp-manager"
+	DeploymentName       = "sap-btp-operator-controller-manager"
+	OperandConfigMapName = "sap-btp-operator-config"
```
jaroslaw-pieszka: Can these two variables be overwritten either by CLI or by ConfigMap?

Afaik these are constants:
```go
	operandName  = "sap-btp-operator"

	sapBtpServiceOperatorClusterIdSecretName = operandName + "-clusterid"
	sapBtpServiceOperatorConfigMapName       = operandName + "-config"
```

controllers/btpoperator_controller_configuration_test.go
```diff
@@ -37,7 +37,18 @@ var _ = Describe("Configuration controller", func() {
 					EnableLimitedCacheConfigMapKey: "false",
 				},
 			}
-			Eventually(func() error { return k8sClient.Create(ctx, configMap) }).Should(Succeed())
+			Eventually(func() error {
```
jaroslaw-pieszka: imho both implementations are not correct.
We should do the action and then call eventually to assert periodically.
So we need to call Create, and then call Get in Eventually.

jaroslaw-pieszka (dismissed):

jaroslaw-pieszka (approved):

PR #1449, opened 2026-04-24
Refactor ymlutils package using yaml.v3 with IO abstraction layer

jaroslaw-pieszka:
## Summary

- Replace manual string-splitting YAML parsing in `internal/ymlutils` with `gopkg.in/yaml.v3`
- Fix colon-split bugs in `ExtractStringValueFromYamlForGivenKey` and `ExtractGvkFromYml`
- Introduce pure-core / OS-wrapper split for single-file functions and `fs.FS` abstraction for directory walkers, enabling unit tests without disk access

## Test plan

- [ ] New unit tests in `internal/ymlutils/ymlutils_test.go` using inline YAML strings and `fstest.MapFS`
- [ ] Existing integration tests in `controllers/` pass unchanged
- [ ] `make test` passes


controllers/btpoperator_controller_updating_test.go
```diff
@@ -124,6 +124,44 @@ var _ = Describe("BTP Operator controller - updating", func() {
 		})
 	})
 
+	When("update some resources names and bump chart version", Label("test-update"), func() {
```
piotrek1459, line 131: This When and It is the same as in line 165. Shoudn't we distinguish them?

controllers/btpoperator_controller_updating_test.go
```diff
@@ -189,7 +227,7 @@ var _ = Describe("BTP Operator controller - updating", func() {
 			err = moveOrCopyNFilesFromDirToDir(len(allManifests), true, getApplyPath(), getTempPath())
 			Expect(err).To(BeNil())
 
-			remainingManifestsNum := 4
+			remainingManifestsNum := 5
```
piotrek1459: I think, it should be enhanced in order to remove "magic number" but this is temporary, right?

controllers/btpoperator_controller_updating_test.go
```diff
@@ -124,6 +124,44 @@ var _ = Describe("BTP Operator controller - updating", func() {
 		})
 	})
 
+	When("update some resources names and bump chart version", Label("test-update"), func() {
+		It("all applied resources should receive new chart version, resources with new names should replace the ones with old names", func() {
+			updateManifestsNum := 3
```
piotrek1459: Here is also "Magic number" maybe we can add some comment explaining why or figure out different approach.

internal/ymlutils/renamer.go
```diff
@@ -1,84 +1,153 @@
...
-	err = os.WriteFile(filename, []byte(output), 0700)
+	updated, err := updateChartVersionInContent(data, newVersion)
 	if err != nil {
 		return err
 	}
-	return nil
+	return os.WriteFile(filename, updated, 0644)
+}
+
+func updateChartVersionInContent(data []byte, newVersion string) ([]byte, error) {
```
piotrek1459, line 146: Is this function used only in tests? It can delete comments or change the order of fields. Is this intended?
if it is for testing purposes I suppose it is ok.

internal/ymlutils/renamer.go
```diff
@@ -1,84 +1,153 @@
...
+}
+
+// addSuffixToNameInContent processes a potentially multi-document YAML byte slice,
+// appending suffix to metadata.name and spec.group in every document,
+// skipping documents whose metadata.name is in excluded.
+func addSuffixToNameInContent(data []byte, suffix string, excluded map[string]bool) ([]byte, error) {
+	dec := yaml.NewDecoder(bytes.NewReader(data))
+	var buf bytes.Buffer
+	enc := yaml.NewEncoder(&buf)
+	enc.SetIndent(4)
```
piotrek1459, line 49: Default for Kubernetes is 2 I think. I know that we do not check raw yaml and those copies are deleted but maybe using 2 would be better for the future if there will be different test cases.

piotrek1459 (approved):
/lgtm

PR #1451, opened 2026-04-25
Simplify setConfigMapValues using a slice of key-value entries

jaroslaw-pieszka:
## Summary
- Replace four repeated `SetNestedField` error-check blocks with a single loop over a slice of key-value pairs
- Eliminates repetition while preserving stable iteration order

## Test plan
- [ ] `make build` passes
- [ ] Existing controller tests pass


piotrek1459 (approved):
/lgtm

PR #1453, opened 2026-04-25
Simplify preSoftDeleteCleanup by deleting directly without prior Get

jaroslaw-pieszka:
## Summary
- Replace three identical Get-then-Delete blocks with a single loop over a slice of `client.Object`
- `r.Delete` returns `NotFound` when the object is absent; `client.IgnoreNotFound` already handles that, so the prior `Get` is unnecessary

## Test plan
- [ ] `make build` passes
- [ ] Existing controller tests pass


piotrek1459 (approved):
/lgtm

PR #1520, opened 2026-06-23
Add Release Management Issue Comment After Channel Promotion

KsaweryZietara:
**Description**

- Add a new step to `promote_module_to_channel.yaml` that comments on the corresponding release-management issue after a patch release is promoted to a channel
- Expose the module-manifests PR URL as a step output so the comment step can reference it
- Skip the comment for new feature releases (patch version 0)

**Related issue(s)**
See also https://github.com/kyma-project/kyma-environment-broker/issues/3294

### DIFF
--- a/.github/workflows/promote_module_to_channel.yaml
+++ b/.github/workflows/promote_module_to_channel.yaml
@@ -80,6 +80,7 @@ jobs:
           git push origin ${BRANCH_NAME}
 
       - name: Create PR if needed
+        id: create_pr
         working-directory: module-manifests
         shell: bash
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/promote_module_to_channel.yaml
```diff
@@ -94,15 +95,56 @@ jobs:
...
-          printf '%s\nRisk is Low\n' "${BTP_MANAGER_RELEASES_URL}/${TAG}" \
-            | gh pr create \
-            -B main \
-            -H ${BRANCH_NAME} \
-            -R "${MODULE_MANIFESTS_REPO_URL}" \
-            --title "Promote BTP Manager ${TAG} to ${CHANNEL} channel" \
-            --fill \
-            --body-file - \
-            | tee $GITHUB_STEP_SUMMARY
+          if [ "${MINOR}" = "3" ]; then
```
jaroslaw-pieszka, line 133: Why do we have hard-wired MINOR in script?

KsaweryZietara: For minor version 3, our PO had to create the microdelivery issue with the title `kyma-project.io/module/btp-operator v1.3.1` (see [#103](https://github.tools.sap/kyma/release-management/issues/103)) instead of `kyma-project.io/module/btp-operator v1.3.0`, because a bug was released in **v1.3.0** and fixed in the **v1.3.1** hotfix.

Future microdelivery issues should use **v1.4.0**, **v1.5.0**, and so on, hopefully.

jaroslaw-pieszka (approved):

PR #1529, opened 2026-06-25
Add CA Bundle Probe v2 E2E Test Infrastructure

jaroslaw-pieszka:
**Description**

- Add fake rt-bootstrapper mutating webhook (`tests/webhook/`) that injects `rt-bootstrapper-certs` volume mount at `/etc/ssl/certs` into pods in `kyma-system` when a `ca-bundle` Secret exists
- Add fake HTTPS server (`tests/fake-server/`) that simulates the token URL endpoint with hot-reloadable TLS cert
- Add probe Job binary (`tests/probe/`) that collects CA bundle mount presence, SHA256 hash, and TLS dial result, then writes results as annotations on the BtpOperator CR — to be spawned dynamically by btp-manager in a subsequent PR
- Add cert generation script (`tests/probe/setup-certs.sh`) producing ca-A, ca-B, and server keypairs for test phases; exports webhook private key to enable cert caching across runs
- Add E2E test script (`scripts/testing/run_e2e_ca_bundle_probe_test.sh`) covering all 7 rows of the MVP decision table across 5 phases (A–E); script only orchestrates world state changes and asserts outcomes — it does not run the probe or touch annotations directly
- Add standalone simulation script (`scripts/testing/simulate_btp_manager.sh`) that stands in for the not-yet-implemented btp-manager probe loop: runs the probe Job, reads BtpOperator CR annotations, restarts sap-btp-operator on CA hash changes, and increments `tls-probe-run-id` so the E2E script can synchronise with it; supports single-cycle and `--loop` modes
- Add `workflow_dispatch`-only GitHub Actions workflow for on-demand E2E runs

The E2E test and simulate script are designed to be split across two terminals: the simulate loop runs independently at a configurable interval, and the E2E script polls `tls-probe-run-id` to detect when the loop has processed each phase's world state change. The `runProbeJob`/`simulateBtpManagerReconcile` inline calls and the simulate script will be removed once btp-manager implements probe Job creation and annotation-driven reconciliation.

**Related issue(s)**
See also https://github.tools.sap/kyma/backlog/issues/9706

### DIFF
--- a/tests/probe/main_test.go
+++ b/tests/probe/main_test.go
@@ -3,9 +3,11 @@ package main
 import (
 	"context"
 	"crypto/x509"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"strings"
 	"testing"
 
 	btpv1alpha1 "github.com/kyma-project/btp-manager/api/v1alpha1"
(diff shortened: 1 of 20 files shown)


tests/probe/main_test.go
```diff
@@ -0,0 +1,158 @@
...
+func TestLoadConfig_Defaults(t *testing.T) {
+	cfg := loadConfig()
+	assert.Equal(t, "kyma-system", cfg.Namespace)
+	assert.Equal(t, "sap-btp-manager", cfg.TLSSecret)
+	assert.Equal(t, "", cfg.TokenURLOverride)
+}
+
+func TestLoadConfig_EnvOverride(t *testing.T) {
+	os.Setenv("PROBE_NAMESPACE", "test-ns")
+	os.Setenv("PROBE_TOKENURL_OVERRIDE", "https://fake.local/health")
```
KsaweryZietara: You can use `t.Setenv()`, which also handles cleanup automatically.

tests/probe/main_test.go
```diff
@@ -0,0 +1,158 @@
...
+	caFile, err := os.CreateTemp("", "ca-*.crt")
+	require.NoError(t, err)
+	content := []byte("-----BEGIN CERTIFICATE-----\nfake\n-----END CERTIFICATE-----\n")
+	_, err = caFile.Write(content)
+	require.NoError(t, err)
+	caFile.Close()
+	defer os.Remove(caFile.Name())
+
+	// Write a fake mountinfo that lists the CA file's directory as a mountpoint
+	mntDir := strings.TrimSuffix(caFile.Name(), "/"+strings.Split(caFile.Name(), "/")[len(strings.Split(caFile.Name(), "/"))-1])
```
KsaweryZietara:
```suggestion
	filepath.Dir(caFile.Name())
```

KsaweryZietara (approved):

PR #1543, opened 2026-07-09
Add CA Bundle Probe Image to Deployment and Configure Interval via ConfigMap

jaroslaw-pieszka:
**Description**

- Fix \`TestClearBtpOperatorProbeAnnotations_ClearsAll\` to assert annotation keys are absent from the map rather than checking for empty string values
- Change \`waitForLastHashToBe\` timeout to return failure (\`return 1\`) instead of silently succeeding
- Move \`ProbeInterval\` from \`PROBE_INTERVAL\` env var to \`sap-btp-manager\` ConfigMap (default: \`0\`, disabled); add \`-probe-interval\` CLI flag
- Add \`PROBE_IMAGE\` to btp-manager Deployment env vars in \`set_external_images.yaml\`
- Build \`ca-bundle-probe\` image in both release workflows, tagged with the release version; add as required job before \`bump-sec-scanners-config\`
- Update \`Makefile\` \`create-manifest\` to substitute probe image tag using \`PROBE_IMG\`
- Update \`examples/btp-operator-configmap.yaml\` and docs with \`ProbeInterval\`

**Related issue(s)**

### DIFF
--- a/.github/workflows/create-release-with-version-submit.yaml
+++ b/.github/workflows/create-release-with-version-submit.yaml
@@ -84,6 +84,16 @@ jobs:
       context: .
       tags: ${{ inputs.name }}
 
+  build-probe-image:
+    name: Build probe image
+    needs: [validate-release]
+    uses: kyma-project/test-infra/.github/workflows/image-builder.yml@main # Usage: kyma-project/test-infra/.github/workflows/image-builder.yml@main
+    with:
+      name: ca-bundle-probe
+      dockerfile: tests/probe/Dockerfile
+      context: .
+      tags: ${{ inputs.name }}
+
   run-unit-tests:
     name: Unit tests
     uses: "./.github/workflows/run-unit-tests-reusable.yaml"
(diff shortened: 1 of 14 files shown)


IwonaLanger (dismissed):

config/manager/set_external_images.yaml
```diff
@@ -11,3 +11,5 @@ spec:
           env:
             - name: SAP_BTP_SERVICE_OPERATOR
               value: europe-docker.pkg.dev/kyma-project/prod/external/ghcr.io/sap/sap-btp-service-operator/controller:v0.11.6
+            - name: PROBE_IMAGE
+              value: europe-docker.pkg.dev/kyma-project/prod/ca-bundle-probe:latest
```
MarekMichali: We shouldn't use latest.

.github/workflows/create-release-with-version-submit.yaml
```diff
@@ -350,7 +360,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          IMG=$IMAGE_REPO:${{ inputs.name }} make create-manifest
+          IMG=$IMAGE_REPO:${{ inputs.name }} PROBE_IMG=europe-docker.pkg.dev/kyma-project/prod/ca-bundle-probe:${{ inputs.name }} make create-manifest
```
MarekMichali: Maybe we could avoid setting this env somehow?

config/manager/set_external_images.yaml
jaroslaw-pieszka: True. I told him that.

Makefile
```diff
@@ -123,6 +124,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 .PHONY: create-manifest
 create-manifest: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
+	sed -i "s|europe-docker.pkg.dev/kyma-project/prod/ca-bundle-probe:[^ ]*|${PROBE_IMG}|g" config/manager/set_external_images.yaml
```
MarekMichali: Why this replacing?

MarekMichali (dismissed):

IwonaLanger (approved):

piotrmiskiewicz (approved):

PR #1581, opened 2026-07-21
Fix CA Bundle Probe Spurious Restarts With Order-Independent Hash

jaroslaw-pieszka:
**Description**

- Replace raw-bytes SHA-256 hash of the CA bundle file with an order-independent hash over per-cert DER fingerprints in `tests/probe/main.go`
- Add `hashCertBundle` helper that parses PEM certs, computes `sha256(cert.Raw)` per cert, deduplicates, sorts fingerprints, and hashes the joined list
- Fix spurious `btp-operator` pod restarts caused by Kubernetes reordering certificates in the mounted volume between writes (same cert set, different byte order → different hash → unnecessary restart)
- Add unit tests covering empty input, single-cert stability, order-independence, different certs produce different hashes, and duplicate cert deduplication

**Related issue(s)**


jaroslaw-pieszka:
Cert expiry fixed — regenerated `certA`/`certB` with 100-year validity (expire 2126). Note that `x509.ParseCertificate` does not validate expiry, so these would have worked indefinitely either way, but the long-lived certs are clearer as stable fixtures.

Regarding the silent empty-hash state: this is by design. When `hashCertBundle` returns `""`, `probe_runner.go` skips the restart check (line 169: `if hash == ""`) and does not advance `tls-probe-last-hash`. A bundle that contains no parseable PEM certificates is treated as "no usable bundle" — and the TLS dial result already covers this case independently: if the cert pool is empty and TLS fails, the probe signals `alert`. Adding a separate error path in `hashCertBundle` for an unreadable bundle would duplicate that signal without adding information.

MarekMichali (approved):

PR #1587, opened 2026-07-23
Run CA Bundle Probe Immediately on Startup

jaroslaw-pieszka:
**Description**

- Run one CA bundle probe cycle immediately when `ProbeRunner.Start` is called, before the first ticker interval fires
- Eliminate the blind window on startup where the probe would not run for a full interval (up to 60 minutes) after btp-manager starts

**Related issue(s)**


jaroslaw-pieszka:
Fixed in the latest commit. After the startup `runCycle` returns an error, we now check `ctx.Err()` and return `nil` immediately if the context is cancelled — consistent with how controller-runtime runnables signal clean shutdown. The ticker loop is not entered in that case.

piotrek1459 (approved):

PR #1617, opened 2026-08-13
Update the troubleshooting guide

IwonaLanger:
**Description**

- Updated `05-04-operator-not-working-after-deleting-binding.md` to more accurately reflect the scenarios users encounter

### DIFF
--- a/docs/user/troubleshooting/05-04-operator-not-working-after-deleting-binding.md
+++ b/docs/user/troubleshooting/05-04-operator-not-working-after-deleting-binding.md
@@ -8,7 +8,7 @@ You deleted the SAP Service Manager service binding in the SAP BTP cockpit and t
 
 When your Kyma runtime was provisioned, an SAP Service Manager service instance with the `service-operator-access` plan and a corresponding service binding were created automatically. The credentials from that binding are used to populate the `sap-btp-manager` Secret in your cluster, which the SAP BTP Operator module relies on to function.
 
-When you delete the binding, the Kyma infrastructure still holds the old credentials and keeps refreshing the `sap-btp-manager` Secret with them. If you re-create the binding, the new binding has different credentials, but the Secret is still not updated. The SAP BTP Operator module can't authenticate, so the entire SAP BTP module remains non-functional until the operations team manually updates the credentials on your behalf.
+When you delete the binding, the Kyma infrastructure still holds the old credentials and keeps refreshing the `sap-btp-manager` Secret with them. If you re-create the binding, the new binding has different credentials, but the Secret is still not updated. You can't create new service instances because the SAP BTP Operator module can't authenticate. The support team must manually update the credentials on your behalf.
 
 > ### Note:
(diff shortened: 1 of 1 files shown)


docs/user/troubleshooting/05-04-operator-not-working-after-deleting-binding.md
```diff
@@ -1,28 +1,39 @@
...
-   - Your subaccount ID
-   - Your Kyma runtime instance ID
+2. In the list of SAP Service Manager, find the service instance whose name matches your Kyma instance ID. This is the dedicated Service Manager instance for your Kyma runtime.
+
+3. In that Service Manager instance, check whether a binding exists.
+   - If no binding exists, re-create it, then proceed to the next step.
+   - If a binding already exists (whether original or re-created), proceed to the next step.
 
-3. Wait for the operations team to confirm that the credentials have been updated and the `sap-btp-manager` Secret on your cluster has been restored. Then verify that the SAP BTP Operator is functioning again by checking that your service instances and bindings are no longer reporting errors.
+4. Check the binding's creation timestamp and compare it with the Service Manager instance's creation timestamp. If the binding was created later, it was deleted and re-created — this confirms you're in the scenario described by this guide.
```
NHingerl:
```suggestion
4. Check the binding's creation timestamp and compare it with the Service Manager instance's creation timestamp:
   - If the binding was created later, it was deleted and recreated. This confirms you're in the scenario described by this guide.
```

to integrate the following "if", maybe make each "if" a list item

docs/user/troubleshooting/05-04-operator-not-working-after-deleting-binding.md
```diff
@@ -1,28 +1,39 @@
...
 
-## Related Information
+   > [!NOTE]
+   > If the timestamps match, the binding has not been re-created and this guide does not apply. The issue may instead be caused by the `sap-btp-manager` Secret being deleted from the cluster. See the [Cause](#cause) section for guidance.
```
NHingerl: Why is one "if" condition ("If the binding was created later" = timestamps mismatch) part of the normal paragraph, but the other "if" condition in a Note? This makes no sense. Either both are nice-to-have info (=> Note) or both are relevant to the main flow (=> normal paragraph). I don't understand the split here.

If in doubt, I'd always prefer a normal paragraph over a note.

```suggestion
   - If the timestamps match, the binding has not been recreated and this guide does not apply. Instead, the issue may be caused by the `sap-btp-manager` Secret being deleted from the cluster. For guidance, see [Cause](#cause).
```

If the reader reads the Cause section again, will this help them solve the issue? If yes, does the Cause section hide an alternative solution to this issue?

IwonaLanger: Yes, it does - it's in line 11. I'll reformat the line as a note for better visibility and have bullet points here.

NHingerl (dismissed):

piotrek1459 (dismissed):
/lgtm

piotrek1459 (approved):

mmitoraj (approved):

grego952 (approved):

Issue #1404, opened 2026-03-17
When modyfing sap-btp-manager configMap with SAP BTP Service Operator settings we need to restart a Pod


PK85:
**Description**

When EnableLimitedCache value is change in the sap-btp-manager ConfigMap (see: https://github.com/kyma-project/btp-manager/blob/main/examples/btp-operator-configmap.yaml), the BTP Manager adjust a configMap for SAP BTP Service Operator but Pod is not restarted and no effect.

**Reason**

Change is not reflected in the cluster.

**AC**
- [x] 1) Any change in the  sap-btp-manager ConfigMap related to the SAP BTP Service Operator component requires pod restart (SAP BTP Service Operator Pod)
  - [ ] 1.1) We plan to refactor how we support  sap-btp-manager ConfigMap and separate configurations for BTP Manager and SAP BTP Service Operator. When it will be done, will be much easier to when to decide to restart pod or not. 
  - [x] 1.2) Without above, we need to hardcode EnableLimitedCache that if chnaged then restart a pod
- [x] 2) Add missing E2E test for EnableLimitedCache changed to false. Right now we have default set to true. Expected test result, pod restart.
- [x] 3) Add documentation for contrubutors how mechanism with restarting pods works, what conditions must be met

PR #1631, opened 2026-08-26
Restart SAP BTP Service Operator pod on EnableLimitedCache change

jeremi-szczotka:
**Description**

- Restart SAP BTP Service Operator pods when `EnableLimitedCache` changes in `sap-btp-manager` ConfigMap — previously the value was updated in `sap-btp-operator-config` but the pod was not restarted, so the change had no runtime effect
- Add envtest test asserting pod deletion is triggered when `EnableLimitedCache` changes
- Add E2E test script `run_e2e_enable_limited_cache_test.sh` that verifies pod restart and config propagation on a real cluster
- Add E2E test step to the main E2E matrix workflow
- Add contributor documentation explaining which config keys require pod restart and why

**Related issue(s)**
#1404


controllers/config/handler.go
```diff
@@ -174,17 +175,38 @@ func (r *Handler) Start(ctx context.Context) error {
...
+	}
+
+	oldEnableLimitedCache := EnableLimitedCache
+	r.applyValues(ctx, cm)
+
+	if EnableLimitedCache != oldEnableLimitedCache {
+		logger := log.FromContext(ctx)
+		if err := r.restartSapBtpServiceOperatorPods(ctx); err != nil {
+			logger.Error(err, "failed to restart SAP BTP service operator pods after EnableLimitedCache change")
+		}
```
MarekMichali: This is a race scenario. We trigger the restart of sap-btp-operator when the btp-manager configmap changes. Before the change is propagated to the sap-btp-operator configmap we could already have restarted the pod without the change.

jeremi-szczotka: Good catch I will fix it.

controllers/config/handler.go
```diff
@@ -63,8 +64,9 @@ type WatchHandler interface {
 
 type Handler struct {
 	client.Client
-	Scheme        *runtime.Scheme
-	configMetrics *metrics.ConfigMetrics
+	Scheme            *runtime.Scheme
+	configMetrics     *metrics.ConfigMetrics
+	pendingPodRestart atomic.Bool
 }
```
MarekMichali: using this flag `pendingPodRestart` looks too complex for an issue like this. Maybe we could just watch the sap-btp-operator-config and react on change?

jeremi-szczotka: Yes. I adjusted solution for this suggestion.

MarekMichali (approved):

PR #1642, opened 2026-09-02
Wait for sap-btp-operator Pod Ready Before Patching ServiceInstance

jeremi-szczotka:
Description: 

- The fix adds a wait loop after ConfigMap propagation that polls until the
`sap-btp-operator` pod is in ready state before continuing. This ensures the webhook is
available when the ServiceInstance patch is attempted.

### DIFF
--- a/scripts/testing/run_e2e_module_tests.sh
+++ b/scripts/testing/run_e2e_module_tests.sh
@@ -391,7 +391,7 @@ ELAPSED=0
 TIMEOUT=120
 while true; do
   READY_COUNT=$(kubectl get pods -n kyma-system -l app.kubernetes.io/instance=sap-btp-operator \
-    -o jsonpath='{range .items[*]}{.status.conditions[?(@.type=="Ready")].status}{"\n"}{end}' 2>/dev/null | grep -c "True" || echo 0)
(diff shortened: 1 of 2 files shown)


scripts/testing/run_e2e_module_tests.sh
```diff
@@ -387,21 +387,9 @@ while true; do
...
-  fi
-  if [[ ${ELAPSED} -ge ${TIMEOUT} ]]; then
-    echo "FAILED: sap-btp-operator pod not Ready within ${TIMEOUT}s" && exit 1
-  fi
-  sleep 5
-  ELAPSED=$((ELAPSED + 5))
-done
+sleep 2
+kubectl wait pod -n kyma-system -l app.kubernetes.io/instance=sap-btp-operator \
+  --for=condition=Ready --timeout=120s
```
MarekMichali: When the pod doesn't restart the test is false positive

MarekMichali: Checking only the status is not enough.

MarekMichali (dismissed):

MarekMichali (approved):

